PR #3, opened 2021-04-17
Goerg model

matthewclegg:
Goerg's single point reach curve model

### DIFF
--- a/src/models/goerg_model.py
+++ b/src/models/goerg_model.py
@@ -1,0 +1,70 @@
+# Copyright 2021 The Private Cardinality Estimation Framework Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


src/models/goerg_model.py
```diff
@@ -0,0 +1,70 @@
...
+      impressions: int, specifies the hypothetical number of impressions that
+        are shown.
+      max_frequency: int, specifies the number of frequencies for which reach
+        will be reported.
+    Returns:
+      An array reach[], where reach[i] is the number of people who would be
+      reached AT LEAST i+1 times.  The length of the array is equal to
+      max_frequency.
+    """
+    kplus_reach_list = []
```
jiayu-google, line 66: We use reach_at_frequency before and kplus_reach_list here. They mean the same thing. I feel it's better to make the name consistent.

src/models/goerg_model.py
```diff
@@ -0,0 +1,70 @@
...
+    for k in range(1,max_frequency+1):
+      kplus_reach = self._rho * (impressions / (impressions + self._beta))**k
```
jiayu-google, line 68: I'm looking at equation (7) of the paper https://storage.googleapis.com/pub-tools-public-publication-data/pdf/43218.pdf

It's different from the formula here. Did I miss something?

jiayu-google (commented):
Thanks Matthew!

jiayu-google (approved):

mpatel24 (approved):

PR #12, opened 2021-05-12
Latin Hypercube test point generator

jiayu-google:
1. Add LatinHypercubeRandomTestPointGenerator.
2. A few other files are affected by applying the Black style.

### DIFF
--- a/src/driver/latin_hypercube_random_test_point_generator.py
+++ b/src/driver/latin_hypercube_random_test_point_generator.py
@@ -59,8 +59,9 @@ class LatinHypercubeRandomTestPointGenerator(TestPointGenerator):
           acceptably small sampling variance for the modeling errors.
         """
         num_points = min(
-          max(self._npublishers ** 2, MINIMUM_NUMBER_OF_TEST_POINTS),
-          MAXIMUM_NUMBER_OF_TEST_POINTS)
+            max(self._npublishers ** 2, MINIMUM_NUMBER_OF_TEST_POINTS),
+            MAXIMUM_NUMBER_OF_TEST_POINTS,
+        )
         design = pyDOE.lhs(n=self._npublishers, samples=num_points, criterion="maximin")
         design = design[self._rng.permutation(num_points), :][
             :, self._rng.permutation(self._npublishers)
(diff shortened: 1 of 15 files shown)


src/driver/latin_hypercube_random_test_point_generator.py
```diff
@@ -0,0 +1,57 @@
...
+        """Returns a generator for generating a list of test points.
+
+        Returns:
+          An iterable of spend vectors representing locations where
+          the true reach surface is to be compared to the modeled reach
+          surface.  A minimum of 100 points will be generated.  This value
+          was chosen heuristically on the belief that this would give an
+          acceptably small sampling variance for the modeling errors.
+        """
+        num_points = max(self._npublishers ** 2, 100)
```
matthewclegg, line 51: Do you want to update this to use a constant as in UniformlyRandomTestPointGenerator?

matthewclegg (changes requested):

src/driver/latin_hypercube_random_test_point_generator.py
```diff
@@ -0,0 +1,70 @@
...
+from typing import Iterable
+from typing import List
+
+import numpy as np
+import pyDOE
+from wfa_planning_evaluation_framework.data_generators.data_set import DataSet
+from wfa_planning_evaluation_framework.driver.test_point_generator import (
+    TestPointGenerator,
+)
+from wfa_planning_evaluation_framework.driver.uniformly_random_test_point_generator import (
```
matthewclegg, line 25: I think the better approach would be to either define a constant MINIMUM_NUMBER_OF_TEST_POINTS that is unique to this class.  Alternatively, define this constant in the base class TestPointGenerator.  Then, both this class and UniformlyRandomTestPointGenerator can import this value from the base class.

matthewclegg (changes requested):

src/driver/latin_hypercube_random_test_point_generator.py
jiayu-google: Makes sense. Now define MINIMUM_NUMBER_OF_TEST_POINTS in the base class TestPointGenerator.

(MAXIMUM_NUMBER_OF_TEST_POINTS is unique to Latin Hypercube)

matthewclegg (approved):

PR #13, opened 2021-05-14
Computes statistics comparing actual vs. estimated reach values

matthewclegg:
Creates a DataFrame of statistics comparing actual reach to estimated reach.

### DIFF
--- a/src/driver/test_point_aggregator.py
+++ b/src/driver/test_point_aggregator.py
@@ -1,0 +1,110 @@
+# Copyright 2021 The Private Cardinality Estimation Framework Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Test Point Aggregator.
+
+Given a collection of true reach values and a corresponding collection
+of reach estimates, computes a single row DataFrame with summary statistics
+on the differences between the true values and the estimated values.
+"""
+
+import numpy as np
+import pandas as pd
+from typing import List
+
+from wfa_planning_evaluation_framework.models.reach_point import ReachPoint
+
+
+# A list of functions for computing aggregation metrics.  Each function
+# takes as input the true reach and modeled reach points and produces
+# as output a scalar value (could be float, int or string).
+AGGREGATORS = {
+    # Number of test points
+    'npoints': lambda x, y: len(x),
+
+    # Mean error (bias)
+    'mean_error': lambda x, y: np.mean(_reach(x) - _reach(y)),
+    
+    # Mean absolute error in predicted reach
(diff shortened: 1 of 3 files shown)


src/driver/test_point_aggregator.py
```diff
@@ -0,0 +1,130 @@
...
+import pandas as pd
+from typing import List
+
+from wfa_planning_evaluation_framework.models.reach_point import ReachPoint
+
+
+# A list of functions for computing aggregation metrics.  Each function
+# takes as input the true reach and modeled reach points and produces
+# as output a scalar value (could be float, int or string).
+AGGREGATORS = {
```
jiayu-google, line 31: Looks like the most compact way of defining a class of methods. Good to learn.

src/driver/test_point_aggregator.py
```diff
@@ -0,0 +1,130 @@
...
+        be of the same length as true_reach.  The value of simulated_reach[i] should be the
+        output of the modeling function for the spend vector that was used to compute
+        true_reach[i].
+    Returns:
+      A single row DataFrame representing the values of the statistics listed in keys.
+    """
+    stats = {}
+    for key in AGGREGATORS:
+        stats[key] = [AGGREGATORS[key](true_reach, simulated_reach)]
+    return pd.DataFrame(data=stats)
```
jiayu-google, line 130: This returns a one-row dataframe with index 0. If we later combine (append) the metrics for N different reach points, we obtain an N-row dataframe with every row index being 0. 

It might be desirable to distinguish each reach point by row index. One option is to add an arg `name` in this method and return `pd.DataFrame(data=stats, index=[name])`. Not sure if it's useful!

src/driver/test_point_aggregator.py
```diff
@@ -0,0 +1,130 @@
...
+    ),
+}
+
+
+def _reach(point_list: List[ReachPoint], k=1) -> np.array:
+    """Returns list of k+ frequencies from list of ReachPoints."""
+    return np.array([point.reach(k) for point in point_list])
+
+
+def _shuffle_distance(xpoint: ReachPoint, ypoint: ReachPoint, k=5) -> float:
```
jiayu-google, line 100: Consider reusing `Simulator._shuffle_distance` in cardinality eval fw: https://github.com/world-federation-of-advertisers/cardinality_estimation_evaluation_framework/blob/4c10fcfd6d687718850067943fdce4bd8c33d6db/src/simulations/simulator.py#L121

jiayu-google (approved):
Thank you Matthew! Left some minor comments.

src/driver/test_point_aggregator.py
matthewclegg: I like this table-driven style because it is very flexible and powerful, although in this case the benefits of using it are not that great.

src/driver/test_point_aggregator.py
matthewclegg: I will keep that in mind for later.  Thanks for the suggestion!

src/driver/test_point_aggregator.py
matthewclegg: I wrote that code!  As a general rule, I think it's a good practice to minimize the number of dependencies between code bases.  Since this implementation computes the shuffle distance between ReachPoints, it wouldn't be possible to eliminate this function altogether.  My feeling was that it was better to reimplement the remainder rather than introduce another dependency.  Other people in the same situation might reasonably choose a different path.

jiayu-google: Thanks for the explanation! I don't have enough experience on this and am very happy to learn your practice and the thought process behind it.

my2457 (approved):

PR #14, opened 2021-05-17
Pairwise union reach surface model

uakyol:


src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,136 @@
...
+    return np.array(res['x']).flatten()
+
+  def get_z_and_alpha(self):
+    """Get intermidiate quantities z and alpha.
+
+    Returns:
+      Tuple of z and alpha.
+      z: Length-n list of length-(p * p) vectors. Each vector is flattened
+      version of the matrix that denotes pairwise overlap regions.
+      alpha: Length-n list of scalars. Each scalar denotes the overlapped reach
```
matthewclegg, line 119: Is alpha the same as b in the doc?

matthewclegg, line 119: non-overlapped?

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,136 @@
...
+        is drawn from ith reach curve.
+    """
+
+    self._reach_curves = copy.deepcopy(reach_curves)
+    self._n = len(reach_points)
+    self._p = len(reach_points[0].impressions)
+    super().__init__(data=reach_points, max_reach=0)
+
+  def _fit(self) -> None:
+    z, alpha = self.get_z_and_alpha()
```
matthewclegg, line 47: Why not use the notation in the "parameter estimation" section of the Pairwise Union Overlap model doc?

src/models/tests/pairwise_union_reach_surface_test.py
```diff
@@ -0,0 +1,114 @@
...
+    max_reaches = [
+        universe_size * (decay_rate**pub_num)
+        for pub_num in range(num_publishers)
+    ]
+    return [
+        LinearCappedReachCurve([ReachPoint([max_reach], (max_reach,))])
+        for max_reach in max_reaches
+    ]
+
+  def genereate_sample_matrix_a(self, num_publishers):
```
matthewclegg, line 74: generate

src/models/tests/pairwise_union_reach_surface_test.py
```diff
@@ -0,0 +1,114 @@
...
+    ]).flatten()
+    for i in range(num_publishers):
+      true_a[i * num_publishers + i] = 0
+    return true_a
+
+  def generate_sample_reach_points(self, true_a, reach_curves, size,
+                                   universe_size):
+    reach_points = []
+    for _ in range(size):
+      impressions = [[np.random.uniform(0, universe_size / 2)]
```
matthewclegg, line 87: In general, it's not a good idea to include randomness in unit tests, because this can make it hard to reproduce the results.

src/models/tests/pairwise_union_reach_surface_test.py
```diff
@@ -0,0 +1,114 @@
...
+
+from absl.testing import absltest
+import numpy as np
+from wfa_planning_evaluation_framework.models.reach_point import ReachPoint
+from wfa_planning_evaluation_framework.models.reach_surface import ReachSurface
+from wfa_planning_evaluation_framework.models.pairwise_union_reach_surface import PairwiseUnionReachSurface
+from wfa_planning_evaluation_framework.models.reach_curve import ReachCurve
+
+
+class LinearCappedReachCurve(ReachCurve):
```
matthewclegg, line 24: You could probably just use GoergModel rather than inventing a new class.

matthewclegg (changes requested):

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,136 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Class for modeling Pairwise union reach surface."""
+
+import copy
+import numpy as np
+from cvxopt import matrix
```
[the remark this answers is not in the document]
uakyol: thank you, I wasn't committing it apparently

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,136 @@
...
+  def get_constraints(self):
+    G1 = -matrix(np.eye(self._p * self._p))
+    G2 = np.kron(np.eye(self._p), np.ones(self._p))
+    G3 = np.kron(np.ones(self._p), np.eye(self._p))
+    h = matrix(
+        np.concatenate((np.zeros(self._p * self._p), np.ones(2 * self._p)),
+                       axis=None))
+    return matrix(np.vstack((G1, G2, G3))), h
+
+  def construct_matrix_a(self):
```
jiayu-google, line 79: This method is also for defining constraints, so better to change the method name, or merge it with the previous method. Explicitly, the previous method gives inequality constraint Ga <= h, while this method gives equality constraint Aa = b (with b = 0 here).

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,136 @@
...
+    reach_sum = sum(reach_vector)
+    overlap = sum([
+        (self._a[i * self._p + j] * reach_vector[i] * reach_vector[j]) /
+        (max(self._reach_curves[i].max_reach, self._reach_curves[j].max_reach) *
+         2) for i in range(self._p) for j in range(self._p)
+    ])
+
+    return ReachPoint(impressions, [reach_sum - overlap])
+
+  def get_constraints(self):
```
jiayu-google, line 70: better to add docstrings

jiayu-google (changes requested):
Thank you Ugur!

My main suggestion is to doc the algorithm description. 3 options off the top of my head:
(1) Describe everything in docstrings. Self-contained but hard to type some formulas. 
(2) Type the formulas (using the corresponding notations like G1) in a markdown, and include the markdown in this repo. Needs additional work.
(3) Type the formulas in LaTeX/doc and upload pdf to this repo.

I'd be happy to discuss and help!

src/models/tests/pairwise_union_reach_surface_test.py
uakyol: I took it from the reach_curve_test. If you don't mind I would like to keep this since it is vert intuitive?
But, I can change if you think there is value in using GoergModel model here.

src/models/tests/pairwise_union_reach_surface_test.py
uakyol: I used a random seed but didn't put impressions as an arg because it is simply generated in this for loop and fits with function name, do you think this works?

src/models/pairwise_union_reach_surface.py
uakyol: ah thank you!

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,136 @@
...
+    G1 = -matrix(np.eye(self._p * self._p))
+    G2 = np.kron(np.eye(self._p), np.ones(self._p))
+    G3 = np.kron(np.ones(self._p), np.eye(self._p))
+    h = matrix(
+        np.concatenate((np.zeros(self._p * self._p), np.ones(2 * self._p)),
+                       axis=None))
+    return matrix(np.vstack((G1, G2, G3))), h
+
+  def construct_matrix_a(self):
```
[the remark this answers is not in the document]
uakyol: yes, sorry I misnamed it. This is the equality constraint matrix, fixed.

src/models/pairwise_union_reach_surface.py
uakyol: no Matthew, z is equivalent to B in the doc.
and d in the doc is ->    d = r_hat(true reach) - alpha

does this make sense? I also outlined how the qp matrices are created in the docstring. If you prefer, I can definitely change the name of the variables.

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,217 @@
...
+        return ReachPoint(impressions, [reach_sum - overlap])
+
+    def get_inequality_constraints(self):
+        """Returns the inequality constraint for the Quadratic Programming solver.
+
+        3 types of inequality constraints are stacked in matrix G and vector h:
+
+          G1: a_i_j >= 0 for all i,j -> all parameters are non negative
+          G2: sum_{j=1..p} a_i_j <= 1 for all i -> row sums are no more than one
+          G3: sum_{i=1..p} a_i_j <= 1 for all j -> col sums are no more than one
```
jiayu-google, line 110: Perhaps clarify the return: 
Returns (matrix G, vector h). Then, cvxopt.solvers recognizes the constraint Gx <= h, where x is the parameter vector.

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,217 @@
...
+            )
+        )
+        return matrix(np.vstack((G1, G2, G3))), h
+
+    def get_equality_constraints(self):
+        """Returns the equality constraint for the Quadratic Programming solver.
+
+        1 type of equality constraint is described by matrix A and vector b:
+
+          a_i_i = 0 for all i -> publishers don't have overlap with themselves.
```
jiayu-google, line 127: Perhaps clarify the return:
Returns (matrix A, vector b). Then, cvxopt.solvers recognizes the constraint Ax = b, where x is the parameter vector.

jiayu-google (approved):
Great, thank you Ugur! Left two minor comments.

src/models/pairwise_union_reach_surface.py
matthewclegg, line 127: Just curious, where in this code are you enforcing the equality constraint that a_{i,j} = a_{j,i}?

src/models/pairwise_union_reach_surface.py
matthewclegg: +1

src/models/tests/pairwise_union_reach_surface_test.py
```diff
@@ -0,0 +1,118 @@
...
+    ]
+
+  def generate_sample_matrix_a(self, num_publishers):
+    true_a = np.array([
+        np.random.dirichlet(np.ones(num_publishers)) * np.random.uniform()
+        for _ in range(num_publishers)
+    ]).flatten()
+    for i in range(num_publishers):
+      true_a[i * num_publishers + i] = 0
+    return true_a
```
matthewclegg, line 84: Does this matrix satisfy the constraint that a[i,j] = a[j,i]?

src/models/tests/pairwise_union_reach_surface_test.py
```diff
@@ -0,0 +1,118 @@
...
+    random_generator = np.random.default_rng(1)
+    for _ in range(size):
+      impressions = [[random_generator.uniform(0, universe_size / 2)]
+                     for _ in range(len(reach_curves))]
+      reach_points.append(
+          self.generate_true_reach(true_a, reach_curves, impressions))
+
+    return reach_points
+
+  def test_by_impressions(self):
```
matthewclegg, line 98: When I run this unit test, I get some additional output, as follows:

```
     pcost       dcost       gap    pres   dres
 0: -3.4733e+09 -3.4733e+09  2e+01  2e+00  3e-10
 1: -3.4733e+09 -3.4733e+09  3e+00  2e-02  3e-12
 2: -3.4733e+09 -3.4733e+09  3e-02  2e-04  3e-14
 3: -3.4733e+09 -3.4733e+09  3e-04  2e-06  3e-16
 4: -3.4733e+09 -3.4733e+09  4e-06  2e-08  6e-17
Optimal solution found.
```

Can you turn this off?

matthewclegg (changes requested):

src/models/pairwise_union_reach_surface.py
jiayu-google: OK, this was not enforced in my colab. So Ugur, please add this constraint (by expanding matrix A).

src/models/tests/pairwise_union_reach_surface_test.py
jiayu-google: Ditto

src/models/tests/pairwise_union_reach_surface_test.py
jiayu-google: Try add `solvers.options['show_progress'] = False` in the beginning

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,217 @@
...
+from typing import Iterable
+from wfa_planning_evaluation_framework.models.reach_point import ReachPoint
+from wfa_planning_evaluation_framework.models.reach_surface import ReachSurface
+from wfa_planning_evaluation_framework.models.reach_curve import ReachCurve
+
+
+class PairwiseUnionReachSurface(ReachSurface):
+    """Models reach with the pairwise union overlap model.
+
+    PairwiseUnionReachSurface models the combined reach with the formula
```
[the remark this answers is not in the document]
jiayu-google: +1

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def _fit(self) -> None:
+        z, alpha = self.get_z_and_alpha()
+        self._a = self.solve_a_given_z_and_alpha(z, alpha)
+
+    def get_reach_vector(self, impressions: Iterable[int]) -> Iterable[int]:
+        """Calculates single publisher reaches for a given impression vector.
+
+        Args:
+          impressions: A list of impressions per publisher. prediction.
```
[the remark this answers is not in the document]
uakyol: sorry, it seems I mistakenly copy/pasted from somewhere

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def __init__(
+        self, reach_curves: Iterable[ReachCurve], reach_points: Iterable[ReachPoint]
+    ):
+        """Constructor for PaiwiseUnionReachSurface.
+
+        Args:
+          reach_curves: A list of ReachCurves to be used in model fitting and reach
+            prediction.
+          reach_points: A list of ReachPoints to which the model is to be fit. This
```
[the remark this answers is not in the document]
uakyol: ah yes, thank you.

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,217 @@
...
+            is parallel to the reach_curves list. The reach point at ith poisition
+            is drawn from ith reach curve.
+        """
+
+        self._reach_curves = copy.deepcopy(reach_curves)
+        self._n = len(reach_points)
+        self._p = len(reach_points[0].impressions)
+        super().__init__(data=reach_points, max_reach=0)
```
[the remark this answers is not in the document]
uakyol: thanks for catching this.

src/models/tests/pairwise_union_reach_surface_test.py
uakyol: thank you! I added at the beginning of the pairwise_union_reach_surface file not the test. What do you think? Do you prefer I add to test or is this good?

matthewclegg: It's better to have it in pairwise_union_reach_surface.py (as it is now).  Thanks!

src/models/tests/pairwise_union_reach_surface_test.py
```diff
@@ -0,0 +1,125 @@
...
```
matthewclegg, line 105: All of this looks good.  I have one additional request, if you don't mind.  Could you please add another two unit tests:

(a) Two reach curves, A and B, with no overlap.  If r_A(i) is the reach curve for A and r_B(i) is the reach curve for B, then the pairwise union model should satisfy r(i_A, i_B) = r_A(i_A) + r_B(i_B).

(b) Two reach curves, A and B, where the viewers in each curve are chosen uniformly at random from a universe of size N.  In this case, the pairwise union model should satisfy r(i_A, i_B) = r_A(i_A) + r_B(i_B) - r_A(i_A) * r_B(i_B) / N.

Thanks for coding all of this.  It looks great!

matthewclegg (changes requested):

src/models/pairwise_union_reach_surface.py
uakyol: closing this based on offline conversation

src/models/tests/pairwise_union_reach_surface_test.py
uakyol: thank you Matthew, I added those test cases too. Hope these make sense.

matthewclegg (approved):

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,251 @@
...
+                constraint_row += 1
+        b = matrix(
+            np.concatenate(
+                (np.zeros(self._p), np.zeros(num_symmetry_constraints)),
+                axis=None,
+            )
+        )
+        return matrix(np.vstack((A1, A2))), matrix(b)
+
+    def solve_a_given_z_and_alpha(self, z: List[List[int]], alpha: List[List[int]]):
```
matthewclegg, line 167: Ditto.

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,251 @@
...
+            sum((self._data[k].reach() - alpha[k]) * z[k] for k in range(self._n))
+        )
+
+        G, h = self.get_inequality_constraints()
+        A, b = self.get_equality_constraints()
+        res = solvers.qp(P, q, G, h, A, b)
+
+        return np.array(res["x"]).flatten()
+
+    def get_z_and_alpha(self):
```
matthewclegg, line 215: Ditto.

matthewclegg (changes requested):

src/models/pairwise_union_reach_surface.py
```diff
@@ -0,0 +1,251 @@
...
+
+        Args:
+          impressions: A list of impressions per publisher.
+
+        Returns:
+          A list R of length p. The value R[i] is the reach achieved on
+          publisher i when impressions[i] impressions are shown.
+        """
+        return [
+            reach_curve.by_impressions(impression).reach()
```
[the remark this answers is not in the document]
jiayu-google: +1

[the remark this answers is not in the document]
uakyol: thank you for finding this. Should be fixed now. I also pulled your changes and applied the fix version and confirmed the test you wrote for M3 worked. Hope it works when these are propagated up the chain too.

matthewclegg (approved):
Thanks, Ugur!

AaronRanAn:
@uakyol thank you for putting this together! I start to go through the outstanding PRs this week (quite slowly due to the sheer volume). 

I reference the doc about the pairwise union model and believe this PR implements it correctly. However, I am not really familiar with how the code is set up and hope you can help provide a test case to let me test a simple case of Pairwise model run?  After that, I am comfortable we merge this one in. 

Thanks!

AaronRanAn (approved):
Feel free to move forward if not blocked by other reviewers.

uakyol:
Thank you so much Aaron! I was thinking perhaps an explanation of the unit tests would be useful. The main one here is "test_by_impressions" in it
1) We generate simulated reach curves - one for each publisher - and with decaying universe sizes.
2) We generate a random matrix A - assuming the generating process for reach is actually a pairwise union model- we generate the model parameters
3) then, using these two, we can generate the reach points. 
4) We then fit the model - since the generating process is the same model, the fitted model should have a really small error rate.

What do you think?

AaronRanAn:
Glad I can help @uakyol! Yeah, knowing the main unit test with the description make a lot more sense for me now. :) 

If you don't mind, we could add those in before you merge. Thanks again!

PR #17, opened 2021-05-18
Smart Experimental Data Design Config

uakyol:


uakyol:
LHS implementation, as well as other pyDOE implementations do not expect a random seed, thus I just tested the number of samples generated rather than the names since it is random. Here is the underlying code: https://github.com/tisimst/pyDOE/blob/master/pyDOE/doe_lhs.py#L21

jiayu-google (commented):
Hi Ugur,

Do you mind adding doctrings for each method now? They can strongly help me understand the logics. Or if you prefer adding them later, could we have a meeting to walk me through the meanings of some arg/return. PLMK!

jiayu-google (changes requested):
Thanks Ugur! Have some small suggestions

src/data_generators/test_lhs_synthetic_data_design_config.py
```diff
@@ -0,0 +1,111 @@
...
+    return [
+        cls.get_data_set_params(
+            cls.get_params_dict(lhs_trial, grid_axes), random_generator)
+        for lhs_trial in lhs_design
+        for i in range(NUM_RANDOM_REPLICAS)
+    ]
+
+  @classmethod
+  def get_data_set_params(cls, params_dict,
+                          random_generator: Generator) -> DataSetParameters:
```
jiayu-google, line 100: Describe the input output, better with an example. Or add a test to show example

src/data_generators/test_rounded_lhs_synthetic_data_design_config.py
```diff
@@ -0,0 +1,90 @@
...
+
+NUM_SAMPLES_FOR_LHS = 3
+NUM_RANDOM_REPLICAS = 2
+
+
+class TestRoundedLHSSyntheticDataDesignConfig(LHSSyntheticDataDesignConfig):
+    """Generates a DataDesign using LHS from a synthetic grid of integers."""
+
+    @classmethod
+    def get_rounded_lhs_design(cls, variable_grid_axes: List[float]):
```
jiayu-google, line 56: I feel this method should be in the base class. Like the M3 design will also use get_rounded_lhs_design. WDYT?

jiayu-google (changes requested):
Thank you for the quick action Ugur! Just one comment.

src/data_generators/test_rounded_lhs_synthetic_data_design_config.py
uakyol: ah I see. Moved it, let me know if this works

jiayu-google (approved):
Great! Thanks for the quick action.

AaronRanAn:
@uakyol the code structure makes sense to me and parameters seem reasonable. However, I would still prefer a simple test case based on this PR so I can dig more. 

Apologies that this PR has not being reviewed for a while, I tried to modified the simple test in the readme.md but you've made a few changes since then. Can you share a simple test case here, then I can probably update the ready me as well. 

Thanks!

AaronRanAn (approved):
spent some additional time on this and think we can move forward with this. In the future instances, maybe we could make the unit test a bit more obvious and easy to follow.

src/data_generators/test_rounded_lhs_synthetic_data_design_config.py
```diff
@@ -0,0 +1,99 @@
...
+    @classmethod
+    def get_lhs_design(cls, variable_grid_axes: Dict[str, List[float]]):
+        return cls.get_rounded_lhs_design(variable_grid_axes)
+
+    @classmethod
+    def get_grid_axes(cls) -> Dict[str, List[int]]:
+        grid_axes = OrderedDict()
+        grid_axes[NUM_PUBS] = [1, 2, 5]
+        grid_axes[LARGEST_PUB_SIZE] = [100, 1000]
+        grid_axes[PUB_RATIO] = [1, 0.5, 0.3, 0.1]
```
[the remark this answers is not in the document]
uakyol: @AaronRanAn thanks a lot for the comments and approving and sorry for not getting back to you earlier. I'll make sure to have more comments on unit tests and make them cleaner in the future. 
And yes, these are those experimental parameters.

PR #26, opened 2021-06-09
Proposed M3 modeling strategy.

matthewclegg:


src/simulator/tests/m3_strategy_test.py
```diff
@@ -0,0 +1,90 @@
...
+        budget: PrivacyBudget,
+        privacy_budget_split: float = 0.5,
+        max_frequency: int = 1,
+    ) -> ReachPoint:
+        rp1 = self.curve.by_spend([spends[0]], max_frequency=max_frequency)
+        rp2 = self.curve.by_spend([spends[1]], max_frequency=max_frequency)
+
+        imps = rp1.impressions[0] + rp2.impressions[0]
+        spend = rp1.spends[0] + rp2.spends[0]
+        freqs = [rp1.reach(i) + rp2.reach(i) for i in range(1, max_frequency + 1)]
```
jiayu-google, line 60: So this fake surface is in fact a non-overlap surface, i.e., a plane. LMK if I understand correctly.

jiayu-google (approved):
Thank you Matthew! It is quite clear and I didn't catch any mistake.

src/simulator/tests/m3_strategy_test.py
matthewclegg: It's true that the audiences for the two reach curves are non-overlapping, but the surface is not a plane.

jiayu-google: You are right.

jiayu-google (approved):

RonPinelliMRC (approved):

PR #29, opened 2021-06-12
Updates Heterogeneous and Heavy Tailed data generators to use numpy.random.Generator

matthewclegg:


src/data_generators/tests/heavy_tailed_impression_generator_test.py
```diff
@@ -23,10 +22,10 @@
 
 class ImpressionGeneratorTest(absltest.TestCase):
     def test_heterogeneous_impression_generator(self):
-        generator = HeavyTailedImpressionGenerator(5, 2, RandomState(0))
+        generator = HeavyTailedImpressionGenerator(5, 2)
         generated_ids = generator()
         self.assertEqual(set(generated_ids), set([0, 1, 2, 3, 4]))
-        self.assertLen(generated_ids, 33)
+        self.assertLen(generated_ids, 13)
```
riemanli, line 28: How do we get the number `13` here? Is it solely based on the result of seed==1?

src/data_generators/heterogeneous_impression_generator.py
```diff
@@ -29,7 +29,7 @@ def __init__(
         n: int,
         gamma_shape: float = 1,
         gamma_scale: float = 1,
-        random_state: RandomState = None,
+        random_generator: np.random.Generator = None,
     ):
         """Constructor for the HeterogeneousImpressionGenerator.
```
riemanli, line 35: Not related to the changes. I think the term in exp() in the pdf of the Gamma distribution in the docstring should have a negative sign.

jiayu-google: Fixed in PR#37. Thanks Rieman.

src/data_generators/tests/heavy_tailed_impression_generator_test.py
jiayu-google: Yes. Result changes when using a different seed.

(I wrote this test before and now realized that a test based on random seed is not a good practice. Let's keep it for now and I'll use "fake seed" for test in the future.)

jiayu-google (approved):
Thanks Matt!

riemanli (approved):
LGTM

mpatel24 (approved):

RonPinelliMRC (approved):

PR #35, opened 2021-06-14
Changes campaign_spends to campaign_spend_fractions in SystemParameters

matthewclegg:
The SystemParameters tuple is intended to specify configurations that will apply across a collection of data sets.  Since different data sets may have campaigns with different maximum spend amounts, it makes sense to express the campaign spend as a fraction of the total possible spend.


src/simulator/system_parameters.py
```diff
@@ -40,20 +40,21 @@ class SystemParameters(NamedTuple):
     """Parameters defining a simulation run.
 
     Attributes:
-        campaign_spend:  A list of spend values, one per campaign,
-            representing the total actual spend for each campaign.
+        campaign_spend_fraction:  A list of values, each between 0 and 1,
```
jiayu-google, line 43: fractions.

src/simulator/system_parameters.py
```diff
@@ -40,20 +40,21 @@ class SystemParameters(NamedTuple):
...
+            one per campaign, representing the amount spent on the campaign
+            as a fractrion of total possible.
```
jiayu-google, line 45: fractrion -> fraction

src/simulator/system_parameters.py
```diff
@@ -40,20 +40,21 @@ class SystemParameters(NamedTuple):
     """Parameters defining a simulation run.
 
     Attributes:
-        campaign_spend:  A list of spend values, one per campaign,
-            representing the total actual spend for each campaign.
+        campaign_spend_fraction:  A list of values, each between 0 and 1,
+            one per campaign, representing the amount spent on the campaign
```
jiayu-google, line 44: I would call out that len(campaign_spend_fractions) = number of publishers

src/driver/sample_experimental_design.py
```diff
@@ -51,9 +51,8 @@
 ]
 
 # Here we assume all data sets specify exactly two campaigns
```
jiayu-google, line 53: I know that exactly two campaigns = exactly two publishers. But "all data sets specify exactly two campaigns" sounds really awkward to me... Just saying.

src/driver/sample_experimental_design.py
```diff
@@ -51,9 +51,8 @@
...
-# of up to $1M each.
-CAMPAIGN_SPENDS = list(
-    itertools.product(np.arange(1, 10) * 100_000, np.arange(1, 10) * 100_000)
+CAMPAIGN_SPEND_FRACTIONS = list(
+    itertools.product(np.arange(1, 10) * 0.1, np.arange(1, 10) * 0.1)
 )
```
jiayu-google, line 57: The length of `CAMPAIGN_SPEND_FRACTIONS` list is num_publishers, which is specified by a dataset.

But, `CAMPAIGN_SPEND_FRACTIONS` is included in `SystemParameters`. There is then a dependency of `SystemParameters` on `DataDesign`, which causes trouble. Am I correct?

src/driver/sample_experimental_design.py
matthewclegg: Could you suggest wording that you think would be less awkward?

src/driver/sample_experimental_design.py
matthewclegg: I agree that there is a problem, but I don't know if this is the best solution.  Yes, a DataSetParameters object does specify a number of publishers.  Since a DataSetParameters object only specifies one DataSet within a DataDesign, data designs can include data sets with differing numbers of publishers.  Also, there is no linkage between any particular data design and any particular experiment design.  One reason for this is that we would like to be able to run multiple different experiments against a single data design, and also we would like to be able to run experiments against data designs from other data sources (e.g., publisher-provided internal data).  So, I don't want to introduce a dependency between the data design and the experimental design.  

What can we do about it?  Here are some options that come to mind:

(1) When an ExperimentalTrial is constructed, check to see if the number of publishers in the DataSet matches the number of publishers in the TrialDescriptor.  If they don't match, then skip that Trial.  In this way, we would only perform experiments when the TrialDescriptor is compatible with the underlying data.

(2) Instead of explicitly specifying the campaign spend fractions, specify a method for generating campaign spend fractions.  The actual vector of spend fractions would be generated on the fly when the Trial Descriptor and the DataSet are joined to create an ExperimentalTrial.

For now, I would propose that we stick with the current design, but add this as a TODO item to be implemented later when we have come up with a better approach.

src/simulator/system_parameters.py
matthewclegg: Done.

src/simulator/system_parameters.py
matthewclegg: Done.

src/simulator/system_parameters.py
matthewclegg: Done.

src/driver/sample_experimental_design.py
jiayu-google: Thanks for the deep thoughts, Matthew! I like both options, and prefer (2) better. Yes, will add this as a TODO item in `m3_first_round_experimental_design`. And will get around it if we encounter any issue when putting all pieces together and do the first round eval.

src/driver/sample_experimental_design.py
jiayu-google: Off the top of my mind.
Option 1. "all data sets include exactly two publishers". Option 2. "all data sets include exactly two publishers and specify a campaign on each publisher".
Reason: when we say "specify two campaigns", it can be the case where we have 3 pubs A, B, C but only launch campaigns on A & B.

jiayu-google (approved):
Thanks Matt!

riemanli (approved):
LGTM.

RonPinelliMRC (approved):

PR #36, opened 2021-06-14
Proposed single publisher experimental design.

matthewclegg:


jiayu-google (approved):
LGTM!

src/driver/experimental_trial.py
```diff
@@ -125,9 +125,9 @@ def evaluate(self, rng: np.random.Generator) -> pd.DataFrame:
             halo, self._trial_descriptor.system_params, privacy_budget
         )
 
-        test_points = self._trial_descriptor.experiment_params.generate_test_points(
+        test_points = list(self._trial_descriptor.experiment_params.generate_test_points(
```
riemanli, line 128: https://github.com/world-federation-of-advertisers/planning-evaluation-framework/blob/26ef53422c3ab83b6336cd656643468770536998/src/driver/experiment_parameters.py#L67

If ```generate_test_points``` doesn't return List, the above typing description should be modified.

src/simulator/single_publisher_strategy.py
```diff
@@ -0,0 +1,67 @@
...
+from wfa_planning_evaluation_framework.simulator.halo_simulator import HaloSimulator
+from wfa_planning_evaluation_framework.simulator.modeling_strategy import (
+    ModelingStrategy,
+)
+from wfa_planning_evaluation_framework.simulator.privacy_tracker import PrivacyBudget
+from wfa_planning_evaluation_framework.simulator.system_parameters import (
+    SystemParameters,
+)
+from wfa_planning_evaluation_framework.simulator.privacy_tracker import (
+    PrivacyBudget,
```
riemanli, line 30: Is there any particular reason to import `PrivacyBudget` twice?

src/driver/experimental_trial.py
matthewclegg: Thanks for pointing this out.  I changed the signature to return Iterable[ReachPoint].

src/simulator/single_publisher_strategy.py
matthewclegg: No.  Thanks for catching that.

riemanli (approved):

RonPinelliMRC (approved):

PR #38, opened 2021-06-17
Add m3 data design following Matthew's example.

jiayu-google:


matthewclegg (changes requested):
Thanks, Jiayu!  Could you please add a test to `tests/synthetic_data_design_generator_test.py` similar to the current test named `test_lhs_design` and make sure that it runs correctly?  (Only 5 lines of code).  Otherwise, it looks good!

jiayu-google:
Thank Matt! Sorry for not having test before. Now refactored the codes, so could you please review again?

src/data_generators/tests/synthetic_data_design_generator_test.py
```diff
@@ -40,6 +46,32 @@ def test_lhs_design(self):
         )
         self.assertLen(list(lhs_design), 10)
 
+    def test_m3_design_size(self):
+        m3_design = m3_data_design.generate_data_design_config(
+            np.random.default_rng(seed=1)
+        )
+        self.assertLen(list(m3_design), 100)
+
+    def test_m3_design_generate_universe_size(self):
```
matthewclegg, line 55: What does this test that is not tested by `test_m3_design_size`?  Is the additional complexity worth it?

jiayu-google: Compared to your `lhs_data_design_example.py`, I added a few lines to `generate_data_design_config()` to correctly specify `universe_size` in `IndependentOverlapDataaSet`. See lines 183-192 of `m3_data_design.py`.

I felt that these lines are slightly complicated and are worth testing, so added this test. It is intended to test if the `IndependentOverlapDataaSet`, in particular, the `universe_size` is specified as expected.

It's indeed worth discussing, please let me know your recommendation, and I'm Ok with either removing or keeping it.

matthewclegg: I think a more succinct approach would be to use unittest.mock to patch the LEVELS dict.

matthewclegg: I think a more succinct approach would be to use unittest.mock to patch the LEVELS dict.  Would you like to try that?

matthewclegg (changes requested):

src/data_generators/tests/synthetic_data_design_generator_test.py
jiayu-google: Sorry, I just saw your response here. Started trying unittest.mock

riemanli (approved):
LGTM

matthewclegg (approved):

RonPinelliMRC (approved):

PR #40, opened 2021-06-21
Add npoints generator

jiayu-google:
Number of test points should increase as number of publishers increases. Added arguments to reflect this.

### DIFF
--- a/src/driver/latin_hypercube_random_test_point_generator.py
+++ b/src/driver/latin_hypercube_random_test_point_generator.py
@@ -50,9 +50,9 @@ class LatinHypercubeRandomTestPointGenerator(TestPointGenerator):
           dataset:  The DataSet for which test points are to be generated.
           rng:  A numpy Generator object that is used to seed the generation
             of random test points.
-          npoints_generator: a function from the number of publishers to the
-            number of test points. EΩxample is
-            lambda npublishers: 2 ** npublishers
+          npoints_generator: A function that returns the number of test points
+            that should be generated. The argument of the function is the number
+            of publishers. Example is `lambda npublishers: 2 ** npublishers`.
             If the function is an empty function, then obtain the number of test
             points directly from the next argument.
           npoints: Number of points to generate.
--- a/src/driver/latin_hypercube_random_test_point_generator.py
+++ b/src/driver/latin_hypercube_random_test_point_generator.py
@@ -15,6 +15,7 @@
 
 from typing import Iterable
 from typing import List
+from typing import Callable
 
 import numpy as np
 import pyDOE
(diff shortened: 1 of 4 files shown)


src/driver/latin_hypercube_random_test_point_generator.py
```diff
@@ -48,11 +50,19 @@ def __init__(
           dataset:  The DataSet for which test points are to be generated.
           rng:  A numpy Generator object that is used to seed the generation
             of random test points.
+          npoints_generator: a function from the number of publishers to the
```
matthewclegg, line 53: Could we change the comment to read, 'A function that returns the number of test points that should be generated.  The argument of the function is the number of publishers.'

src/driver/latin_hypercube_random_test_point_generator.py
```diff
@@ -48,11 +50,19 @@ def __init__(
...
+            number of test points. EΩxample is
```
matthewclegg, line 54: Fix typo in the word 'example'.

matthewclegg (approved):
This seems like a useful change.  Thanks for adding it.  One thing I wonder about is whether it makes sense to get rid of the 'npoints' argument, and instead set the default value for npoints_generator to lambda p: MINIMUM_NUMBER_OF_TEST_POINTS.  I am OK with the PR in its current form, but I thought you might want to consider this.

src/driver/uniformly_random_test_point_generator.py
```diff
@@ -37,19 +38,28 @@ def __init__(
         self,
         dataset: DataSet,
         rng: np.random.Generator,
+        npoints_generator: Callable[[int], int] = lambda x: None,
         npoints: int = MINIMUM_NUMBER_OF_TEST_POINTS,
     ):
-        """Returns a UniformlyRandomTestPointGenerator.
+        """Returns a LatinHypercubeRandomTestPointGenerator.
```
riemanli, line 44: I might misunderstand something, but I feel the original is correct 😅

jiayu-google: Thanks for catching this! I did too many copy pastes 😅

src/driver/latin_hypercube_random_test_point_generator.py
jiayu-google: Thanks for catching this!

src/driver/latin_hypercube_random_test_point_generator.py
jiayu-google: Sure. It does read better.

riemanli (approved):

RonPinelliMRC (approved):

PR #41, opened 2021-06-21
Add campaign_spend_fractions_generator to SystemParameters

jiayu-google:


src/simulator/system_parameters.py
```diff
@@ -40,21 +41,35 @@ class SystemParameters(NamedTuple):
...
-    liquid_legions: LiquidLegionsParameters
-    generator: Generator
+    campaign_spend_fractions: List[float] = [0]
+    liquid_legions: LiquidLegionsParameters = LiquidLegionsParameters()
+    generator: np.random.Generator = np.random.default_rng(1)
+    campaign_spend_fractions_generator: Callable[[int], List[float]] = lambda x: None
 
     def __str__(self) -> str:
         spend_str = ",".join([f"{s}" for s in self.campaign_spend_fractions])
+        # TODO(jiayu): add str for campaign_spend_fractions_generator if necessary
```
matthewclegg, line 72: I don't think this would be necessary, as long as we keep the campaign_spend_fractions in the string.

matthewclegg, line 72: I don't think this would be necessary, as long as we keep the campaign_spend_fractions in the string.

matthewclegg (approved):
(1) Does it make sense to eliminate the campaign_spend_fractions argument?

(2) I wonder if the campaign_spend_fractions_generator should take an numpy.random.Generator argument to allow for the possibility of generating random spend fractions?

src/simulator/system_parameters.py
jiayu-google: makes sense. deleted the todo item.

RonPinelliMRC (approved):

PR #43, opened 2021-06-25
Design for running a quick analysis and for use in Colab

jiayu-google:
The evaluation is too slow even when using `simple_data_design_example` and `sample_experimental_design`. In order to quickly get a meaningful result and establish an analysis practice in Colab, I created some simpler but still meaningful designs.

These simpler designs are simply copied from existing designs with parameters changed. In particular, the reach is decreased to reduce the time of adding ids into the sketches, and other parameters are adjusted according.


src/models/goerg_model.py
```diff
@@ -44,15 +46,30 @@ def __init__(self, data: [ReachPoint]):
...
-            self._impressions - self._reach
-        )
+        if abs(self._impressions - self._reach) < 0.001:
+            # In this corner case, there will be a division by zero error if
+            # we estimate rho using the formula. This error will block the rest
+            # of evaluation. To avoid blocking the rest of evaluation,
+            # we will assign rho a hard-coded maximum value, which is 100 here.
+            # TODO(jiayu): Think about the choice of max value or alternatives.
+            warnings.warn(
+                "impression = reach. rho is assigned a hard-coded maximum value."
```
matthewclegg, line 66: Output the value of impressions and reach in this error message.

src/models/goerg_model.py
```diff
@@ -44,15 +46,30 @@ def __init__(self, data: [ReachPoint]):
         self._fit()
         self._max_reach = self._rho
         if data[0].spends:
-            self._cpi = data[0].spends[0] / data[0].impressions[0]
+            if data[0].impressions[0] < 0.001:
```
matthewclegg, line 49: Should it be considered an error if the Goerg model is called with an impression value that is close to zero?

src/driver/tests/experiment_driver_test.py
```diff
@@ -108,6 +114,28 @@ def test_experiment_driver_with_m3_first_round_experimental_design(self):
...
+                data_design_dir, 1, analysis_example_data_design.__file__, False
+            )
+            data_design_generator()
+            rng = np.random.default_rng(seed=1)
+            experimental_design = analysis_example_experimental_design.__file__
+            experiment_driver = ExperimentDriver(
+                data_design_dir, experimental_design, output_file, intermediate_dir, rng
+            )
+            result = experiment_driver.execute()
+            print(result.shape[0])
```
matthewclegg, line 136: Remove print statement and uncomment assertion.

matthewclegg (changes requested):

src/driver/tests/experiment_driver_test.py
jiayu-google: Done.

PR #2, opened 2021-04-17
Basic model classes -- reach point, reach curve and reach surface

matthewclegg:
Base class for single publisher models

PR #46, opened 2021-07-08
Compute reach in Venn diagram regions

riemanli:
Compute k+ reaches in Venn diagram regions given spends.


src/simulator/halo_simulator.py
```diff
@@ -223,17 +237,59 @@ def simulated_venn_diagram_reach_by_spend(
...
+
+        # Fill in 2^p - 1 regions of the Venn diagram. Note that the region with
+        # representation 0 is unused
+        num_publishers = len(spends)
+        regions = [{} for _ in range(2 ** num_publishers)]
+        for user_id, user_info in user_info.items():
+            lr = user_info.located_region
+            regions[lr][user_id] = user_info.impressions
+
+        return pub_set_by_region, regions
```
riemanli, line 291: @matthewclegg @jiayu-google 
I know we have said to compute the `reach` (number of unique users) in the regions, but I wonder if obtaining more information than just `reach` will be helpful. 

For example, the code above obtains regions such that every region contains `frequency` info represented by a dictionary of `user_id` to number of impressions. This might be useful later on when we compute **impressions of publishers** and **k+ reaches** for initializing a reach point.

If that's not what we want, it's easy to switch back to computing `reach` by counting the number of users in a region.

Please let me know what you think. Thanks!

src/simulator/halo_simulator.py
```diff
@@ -38,6 +40,10 @@
 from wfa_cardinality_estimation_evaluation_framework.estimators.estimator_noisers import (
     GeometricEstimateNoiser,
 )
+from wfa_cardinality_estimation_evaluation_framework.estimators.any_sketch import (
```
matthewclegg, line 43: Is this needed?

src/simulator/halo_simulator.py
```diff
@@ -223,17 +237,59 @@ def simulated_venn_diagram_reach_by_spend(
...
         Returns:
-            A list of ReachPoint. Each reach point represents the mapping from 
+            A list of ReachPoint. Each reach point represents the mapping from
             the spends of a subset of publishers to the differentially private
             estimate of the number of people reached in this subset.
         """
         raise NotImplementedError()
 
+    def _form_venn_diagram_regions(self, spends: List[float]):
+        """Form Venn diagram regions that contain frequency"""
```
matthewclegg, line 253: Add detailed description of arguments.  Also, what is the return value?

src/simulator/halo_simulator.py
```diff
@@ -54,6 +60,14 @@
 )
 
 
+MAX_ACTIVE_PUBLISHERS = 20
+
+
+class UserInfo(NamedTuple):
```
matthewclegg, line 66: Is this being used?

src/simulator/halo_simulator.py
```diff
@@ -223,17 +237,59 @@ def simulated_venn_diagram_reach_by_spend(
...
+        # Fill in 2^p - 1 regions of the Venn diagram. Note that the region with
+        # representation 0 is unused
+        num_publishers = len(spends)
+        regions = [{} for _ in range(2 ** num_publishers)]
+        for user_id, user_info in user_info.items():
+            lr = user_info.located_region
+            regions[lr][user_id] = user_info.impressions
+
+        return pub_set_by_region, regions
+
```
matthewclegg, line 292: Please add some unit tests.

matthewclegg (changes requested):

src/simulator/halo_simulator.py
riemanli: Nope. It's for noise adding. Do it too early. 
Thanks for catching that!

src/simulator/halo_simulator.py
riemanli: Yes. Please see the defaultdict variable `user_info`.

src/simulator/halo_simulator.py
riemanli: Based on the discussion with @matthewclegg , we will only compute `reach` in the regions. As a result, the k+ reaches will be 1+ reaches. The impressions of publishers can be directly computed from the spend vector.

src/simulator/halo_simulator.py
```diff
@@ -223,17 +237,59 @@ def simulated_venn_diagram_reach_by_spend(
...
+        user_counts_by_pub_id = {}
+        for pub_id, spend in enumerate(spends):
+            if not spend:
+                continue
+            user_counts_by_pub_id[pub_id] = self._publishers[
+                pub_id
+            ]._publisher_data.user_counts_by_spend(spend)
+
+        if len(user_counts_by_pub_id) > MAX_ACTIVE_PUBLISHERS:
+            raise ValueError("Too many active publishers for Venn diagram algorithm.")
```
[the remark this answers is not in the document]
riemanli: Done

src/simulator/halo_simulator.py
riemanli: Done

src/simulator/halo_simulator.py
riemanli: Decided to remove `UserInfo` class and create a `collections.namedtuple` inside the function.

src/simulator/halo_simulator.py
```diff
@@ -223,17 +228,98 @@ def simulated_venn_diagram_reach_by_spend(
...
-            A list of ReachPoint. Each reach point represents the mapping from 
+            A list of ReachPoint. Each reach point represents the mapping from
             the spends of a subset of publishers to the differentially private
             estimate of the number of people reached in this subset.
         """
         raise NotImplementedError()
 
+    def _form_venn_diagram_regions(
+        self, spends: List[float], max_frequency: int = 1
+    ) -> Tuple[DefaultDict, List[List]]:
```
matthewclegg, line 245: Suppose there are 50 publishers, so len(spends) = 50, but only 10 nonzero values in spends.  What will be the size of the return value?

matthewclegg (changes requested):

src/simulator/halo_simulator.py
riemanli: For `pub_set_by_region` (`defaultdict`), its size is <= 51, depending on how many regions are occupied.

For `regions` (`List of List`), there are `2^p` lists, and each inner list has `max_frequency` integer values.

riemanli: For `pub_set_by_region` (`defaultdict of `), its size is <= 2^10, depending on how many regions are occupied.
For `regions` (`List of List`), there are 2^50 lists, and each inner list has `max_frequency` integer values. 

Two possible modifications came to me while writing:
1. Instead of output `pub_set_by_region` as `defaultdict`, we can cast it to `dict` to avoid adding pairs that don't exist in the `defaultdict` by querying them.
2. We might not care about the regions that don't have any reach, even they belong to some active publishers. If that's the case, we can use `Dict[int, List]` instead to record the mapping of the representations of occupied regions to k+ frequencies. This will save a lot of space in the case of your example.

Wdyt?

matthewclegg: Yes, I think that #2 would be an improvement.  Would you like to modify your code accordingly?

riemanli: Yes. Going to do it now.

src/simulator/halo_simulator.py
```diff
@@ -223,17 +228,98 @@ def simulated_venn_diagram_reach_by_spend(
...
+
+        if len(user_counts_by_pub_id) > MAX_ACTIVE_PUBLISHERS:
+            raise ValueError(
+                f"There are {len(user_counts_by_pub_id)} publishers"
+                f"for Venn diagram algorithm. The maximum limit is {MAX_ACTIVE_PUBLISHERS}"
+            )
+
+        # Locate user's region and sum the impressions. Also, map the region
+        # representation to publisher ids.
+        UserInfo = collections.namedtuple(
```
matthewclegg, line 280: I think the code would be more compact and probably more readable if you just had two dictionaries, user_region and user_impressions.  User_region would map a user id to the binary representation of the primitive Venn diagram region containing that user, and user_impressions would map a user id to the total number of impressions that had been delivered to that user.

src/simulator/halo_simulator.py
```diff
@@ -223,17 +228,98 @@ def simulated_venn_diagram_reach_by_spend(
...
+
+                users_info[user_id] = UserInfo(
+                    new_lr, users_info[user_id].impressions + impressions
+                )
+
+        # Remove the null region
+        pub_set_by_region.pop(0)
+
+        # Fill in the occupied regions of the Venn diagram with capped user counts.
+        region_user_counts = collections.defaultdict(dict)
```
matthewclegg, line 306: I think lines 306-319 could be replaced with something similar to the following:

frequencies_by_region = [[0] * max_frequency for r in set(user_region.values())]
for user, region in user_region.items():
  frequencies_by_region[region][user_impressions[user]] += 1

... Now you need to convert this into a list of k+ reaches ...

src/simulator/tests/halo_simulator_test.py
```diff
@@ -89,6 +89,88 @@ def test_simulated_reach_by_spend_with_privacy(self, mock_laplace_mechanism):
...
+        expected_pub_set_by_region = {
+            1: set([0]),
+            2: set([1]),
+            3: set([0, 1]),
+        }
+        expected_regions = {1: [1, 1], 2: [1, 0], 3: [0, 1]}
+        pub_set_by_region, regions = halo._form_venn_diagram_regions(spends, max_freq)
+        self.assertEqual(expected_pub_set_by_region, pub_set_by_region)
+        self.assertEqual(expected_regions, regions)
+
```
matthewclegg, line 173: Could you add a unit test with:

pdf1 = PublisherData([(1, 0.01)], "pdf1")
pdf2 = PublisherData([(1, 0.01)], "pdf2")

spends = [0.01, 0.01]

matthewclegg (changes requested):

src/simulator/halo_simulator.py
```diff
@@ -223,17 +228,98 @@ def simulated_venn_diagram_reach_by_spend(
...
+        Args:
+            spends:  The hypothetical spend vector, equal in length to
+              the number of publishers.  spends[i] is the amount that is
+              spent with publisher i. Note that publishers with 0 spends will
+              not be included in the Venn diagram reach.
+            max_frequency:  The maximum frequency for which to report reach.
+        Returns:
+            pub_set_by_region:  Contains the mapping of a binary representation
+              to a set of publisher ids that are in the corresponding region.
+            regions:  Contains k+ frequencies indexed by their binary
```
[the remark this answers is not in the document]
riemanli: Thanks for catching this! I was stupidly mixing frequencies and k+ reaches... my intention was to output frequencies at that moment, but after giving it a second thought, I think outputting k+ reaches makes more sense.
Will modify the docstring as per suggestion.

src/simulator/halo_simulator.py
riemanli: Thanks for the suggestion! Few suggestions on top of yours:
1. Change `frequencies_by_region` to be `Dict[int, List]`, so the indexing later would make sense.
2. Change the length of the lists in `frequencies_by_region` to (max_frequency + 1)
3. For `frequencies_by_region[region][user_impressions[user]] += 1`, `user_impressions[user]` might be larger than `max_frequency`. We can cap it beforehand, ie:
    ```python
     impressions = min(user_impressions[user], max_frequency)
    ```

src/simulator/halo_simulator.py
riemanli: Added.

src/simulator/tests/halo_simulator_test.py
```diff
@@ -89,6 +89,88 @@ def test_simulated_reach_by_spend_with_privacy(self, mock_laplace_mechanism):
...
+        halo = HaloSimulator(data_set, params, privacy_tracker)
+
+        spends = [0.05, 0.08]
+        max_freq = 2
+        expected_pub_set_by_region = {
+            1: set([0]),
+            2: set([1]),
+            3: set([0, 1]),
+        }
+        expected_regions = {1: [1, 1], 2: [1, 0], 3: [0, 1]}
```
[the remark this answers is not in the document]
riemanli: Thanks for catching this. I was stupidly mixing frequencies and k+ reaches... my intention was to test frequencies, but as mentioned in the previous [discussion](https://github.com/world-federation-of-advertisers/planning-evaluation-framework/pull/46/files#r668171952), I will modify the code so the regions will contain k+ reaches instead of frequencies and modify the unit tests accordingly.

src/simulator/halo_simulator.py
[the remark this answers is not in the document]
riemanli: Done

src/simulator/tests/halo_simulator_test.py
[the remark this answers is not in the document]
riemanli: Done

src/simulator/halo_simulator.py
riemanli: Done

src/simulator/halo_simulator.py
riemanli: Done

src/simulator/tests/halo_simulator_test.py
riemanli: Done

src/simulator/halo_simulator.py
```diff
@@ -223,17 +228,94 @@ def simulated_venn_diagram_reach_by_spend(
...
+        user_impressions = collections.defaultdict(int)
+        pub_set_by_region = collections.defaultdict(set)
+
+        for pub_id, user_counts in user_counts_by_pub_id.items():
+            for user_id, impressions in user_counts.items():
+                region = user_region[user_id]
+                new_region = region | (1 << pub_id)
+
+                if not new_region in pub_set_by_region:
+                    pub_set_by_region[new_region] = copy.deepcopy(
```
matthewclegg, line 291: Lines 290-294 could be simplified to:
  pub_set_by_region[new_region] = pub_set_by_region[region] | set(pub_id)

src/simulator/halo_simulator.py
```diff
@@ -223,17 +228,94 @@ def simulated_venn_diagram_reach_by_spend(
...
+        # Get user counts by spend for each active publisher
+        user_counts_by_pub_id = {}
+        for pub_id, spend in enumerate(spends):
+            if not spend:
+                continue
+            user_counts_by_pub_id[pub_id] = self._publishers[
+                pub_id
+            ]._publisher_data.user_counts_by_spend(spend)
+
+        if len(user_counts_by_pub_id) > MAX_ACTIVE_PUBLISHERS:
```
matthewclegg, line 273: Is there a unit test that exercises this?

matthewclegg (changes requested):

src/simulator/tests/halo_simulator_test.py
```diff
@@ -89,6 +89,124 @@ def test_simulated_reach_by_spend_with_privacy(self, mock_laplace_mechanism):
...
+        data_set = DataSet([pdf1, pdf2], "test")
+        params = SystemParameters(
+            [0.4, 0.5], LiquidLegionsParameters(), np.random.default_rng(1)
+        )
+        privacy_tracker = PrivacyTracker()
+        halo = HaloSimulator(data_set, params, privacy_tracker)
+
+        spends = [0.01, 0.01]
+        max_freq = 1
+        expected_pub_set_by_region = {1: set([0]), 3: set([0, 1])}
```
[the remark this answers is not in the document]
riemanli: In the current implementation, when we generate `pub_set_by_region`, we record every combination of publishers along the path to the end. On the other hand, `regions` only records the final regions that contain reach.

Do you want `regions` also records the regions that don't have any reach?

src/simulator/halo_simulator.py
riemanli: No. I will add one. Thanks for pointing out.

src/simulator/halo_simulator.py
riemanli: Good idea. Will do!

src/simulator/tests/halo_simulator_test.py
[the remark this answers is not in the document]
matthewclegg: No, I think both dictionaries should record only regions with positive reach.

src/simulator/halo_simulator.py
riemanli: Done

src/simulator/halo_simulator.py
riemanli: Done

src/simulator/tests/halo_simulator_test.py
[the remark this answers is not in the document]
riemanli: Based on the discussion with @matthewclegg, `pub_set_by_region` is removed.

matthewclegg (approved):

src/simulator/halo_simulator.py
```diff
@@ -223,17 +227,78 @@ def simulated_venn_diagram_reach_by_spend(
...
+        who are reached by the publishers in that subset.
+
+        Args:
+            spends:  The hypothetical spend vector, equal in length to
+              the number of publishers.  spends[i] is the amount that is
+              spent with publisher i. Note that publishers with 0 spends will
+              not be included in the Venn diagram reach.
+            max_frequency:  The maximum frequency for which to report reach.
+        Returns:
+            regions:  Contains k+ reaches indexed by their binary representation.
```
jiayu-google, line 257: How about expanding this sentence:
regions: a dictionary in which each key are the binary representations of each primitive region of the Venn diagram, and each value is a list of the k+ reaches in the corresponding region.

riemanli: Sure. The suggested explanation is more clear.

riemanli: Done

src/simulator/halo_simulator.py
```diff
@@ -223,17 +227,80 @@ def simulated_venn_diagram_reach_by_spend(
...
+                f"for Venn diagram algorithm. The maximum limit is {MAX_ACTIVE_PUBLISHERS}"
+            )
+
+        # Locate user's region represented by number and sum the impressions.
+        user_region = defaultdict(int)
+        user_impressions = defaultdict(int)
+
+        for pub_id, user_counts in user_counts_by_pub_id.items():
+            for user_id, impressions in user_counts.items():
+                user_region[user_id] |= 1 << pub_id
```
jiayu-google, line 284: I like the trick here, and it'll better to explain the expression, like using an example.

jiayu-google (approved):
Great, thank you Rieman!

src/simulator/halo_simulator.py
riemanli: Add comments as per your request. Please let me know if that works for you.

RonPinelliMRC (approved):

PR #47, opened 2021-07-08
Modifies experimental_trial to catch and record exceptions in modeling.

matthewclegg:


src/driver/test_point_aggregator.py
```diff
@@ -124,7 +124,21 @@ def aggregate(
...
     return pd.DataFrame(data=stats)
+
+
+def aggregate_on_failure(inst: Exception) -> pd.DataFrame:
+    """Returns a DataFrame of the same shape as aggregate but for case of failure.
+
+    Args:
+      inst:  The exception instance that was generated in the modeling attempt.
+    Returns:
+      A single row DataFrame representing the values of the statistics listed in keys.
```
jiayu-google, line 139: perhaps say "A single row DataFrame of NAs with columns being the statistics listed in keys".

src/driver/test_point_aggregator.py
```diff
@@ -124,7 +124,21 @@ def aggregate(
...
       A single row DataFrame representing the values of the statistics listed in keys.
     """
-    stats = {}
+    stats = {"model_succeeded": [1], "model_exception": [""]}
     for key in AGGREGATORS:
         stats[key] = [AGGREGATORS[key](true_reach, simulated_reach)]
     return pd.DataFrame(data=stats)
+
+
+def aggregate_on_failure(inst: Exception) -> pd.DataFrame:
```
jiayu-google, line 133: How about calling it `aggregate_on_exception`?  

The essential question: is there any failure (model_succeeded = 0) but non-exception cases? I believe you've thought of them.

Two possible cases off the top of my head: (1) all training points have too small or even negative reach so that the fitted model is not meaningful (2) The optimizer raises a serious warning so that we know the optimization doesn't converge well.

jiayu-google (approved):
Excellent! Thank you Matt.

src/driver/experimental_trial.py
```diff
@@ -121,31 +122,36 @@ def evaluate(self, rng: np.random.Generator) -> pd.DataFrame:
...
             )
-            for t in test_points
-        ]
+            true_reach = [
+                halo.true_reach_by_spend(
+                    t, self._trial_descriptor.experiment_params.max_frequency
+                )
+                for t in test_points
+            ]
+            simulated_reach = [
```
jiayu-google, line 141: perhaps fitted_reach?

src/driver/test_point_aggregator.py
matthewclegg: Yes, that's a better name.

src/driver/test_point_aggregator.py
matthewclegg: That is also better.

src/driver/experimental_trial.py
matthewclegg: Yes

jiayu-google (commented):
LGTM!

RonPinelliMRC:
MRC approving (post merge) based on cursory review after confirming non-independent code review on an exception basis. Experimental code will require independent review prior to publication of or reliance on results generated.

PR #50, opened 2021-07-19
Generate reach points from Venn Diagram

riemanli:


src/simulator/tests/halo_simulator_test.py
```diff
@@ -90,6 +90,93 @@ def test_simulated_reach_by_spend_with_privacy(self, mock_laplace_mechanism):
...
+        },
+        {
+            "testcase_name": "with_1_active_pub_from_2_pubs",
+            "num_publishers": 2,
+            "spends": [0.04, 0.02],
+            "regions": {1: [2]},
+            "expected": [
+                ReachPoint([3, 0], [2], [0.04, 0]),
+                ReachPoint([0, 0], [0], [0, 0.04]),
+                ReachPoint([3, 0], [2], [0.04, 0.04]),
```
jiayu-google, line 121: The spends do not seem correct in here and the case below.
Would you add `assertEqual` for `expected_r_pt.spends`?

src/simulator/tests/halo_simulator_test.py
```diff
@@ -90,6 +90,93 @@ def test_simulated_reach_by_spend_with_privacy(self, mock_laplace_mechanism):
...
+    )
+    def test_generate_reach_points_from_venn_diagram(
+        self, num_publishers, spends, regions, expected
+    ):
+        pdfs = [
+            PublisherData([(1, 0.01), (2, 0.02), (1, 0.04), (3, 0.05)], "pdf1"),
+            PublisherData([(2, 0.03), (4, 0.06)], "pdf2"),
+            PublisherData([(2, 0.01), (3, 0.03), (4, 0.05)], "pdf3"),
+        ]
+        data_set = DataSet(pdfs[:num_publishers], "test")
```
jiayu-google, line 155: Super compact test. Like it.

jiayu-google (changes requested):
Great job!

src/simulator/halo_simulator.py
```diff
@@ -235,6 +235,50 @@ def simulated_venn_diagram_reach_by_spend(
...
+
+        For each subset of active publishers, compute reach and frequency
+        estimate for those users who are reached by at least one of the
+        publishers in the subset.
+
+        Args:
+            spends:  The hypothetical spend vector, equal in length to
+              the number of publishers.  spends[i] is the amount that is
+              spent with publisher i.
+            primitive_regions:  Contains k+ reaches in the regions. The k+
```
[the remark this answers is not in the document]
matthewclegg: +1

src/simulator/halo_simulator.py
```diff
@@ -235,6 +235,50 @@ def simulated_venn_diagram_reach_by_spend(
...
+        impressions = self._data_set.impressions_by_spend(spends)
+
+        reach_points = []
+        for sub_pub_ids in active_pub_powerset:
+            sub_reach = self._aggregate_reach_in_primitive_venn_diagram_regions(
+                sub_pub_ids, primitive_regions
+            )
+            pub_subset = set(sub_pub_ids)
+            sub_imps = [
+                imps if pub_id in pub_subset else 0
```
matthewclegg, line 272: It's slightly more compact to write this as
```
  imps for pub_id, imps in enumerate(impressions) if pub_id in pub_subset
```

Maybe an even more compact approach would be as follows:
```
  pub_vector = np.array([int(i in sub_pub_ids) for i in range(len(spends))])
  sub_imps = np.array(imps) * pub_vector
  sub_spends = np.array(spends) * pub_vector
```

src/simulator/tests/halo_simulator_test.py
```diff
@@ -90,6 +90,93 @@ def test_simulated_reach_by_spend_with_privacy(self, mock_laplace_mechanism):
...
+            "expected": [],
+        },
+        {
+            "testcase_name": "with_1_active_pub_from_2_pubs",
+            "num_publishers": 2,
+            "spends": [0.04, 0.02],
+            "regions": {1: [2]},
+            "expected": [
+                ReachPoint([3, 0], [2], [0.04, 0]),
+                ReachPoint([0, 0], [0], [0, 0.04]),
```
matthewclegg, line 120: Shouldn't the spend vector in this case be [0, 0.02]?

src/simulator/tests/halo_simulator_test.py
```diff
@@ -90,6 +90,93 @@ def test_simulated_reach_by_spend_with_privacy(self, mock_laplace_mechanism):
...
+            "regions": {},
+            "expected": [],
+        },
+        {
+            "testcase_name": "with_1_active_pub_from_2_pubs",
+            "num_publishers": 2,
+            "spends": [0.04, 0.02],
+            "regions": {1: [2]},
+            "expected": [
+                ReachPoint([3, 0], [2], [0.04, 0]),
```
matthewclegg, line 119: The impressions list [3, 0] says that 3 impressions were shown.  But the k+ reach vector of [2] says that only 2 impressions were shown.  One of these two numbers must be incorrect.

matthewclegg (changes requested):

src/simulator/halo_simulator.py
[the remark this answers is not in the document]
riemanli: Thanks for pointing out.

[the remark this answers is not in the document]
riemanli: Enhanced the description of `primitive_regions` in the docstring.

src/simulator/tests/halo_simulator_test.py
riemanli: Thanks for catching that!

src/simulator/tests/halo_simulator_test.py
riemanli: Yes it should. Thanks for pointing out.

src/simulator/tests/halo_simulator_test.py
riemanli: I think they both are correct. Please see the explanation below:
The impressions list shows that given the spends vector = [0.04, 0], ie. only publisher_1, we have 3 impressions for publisher_1 -- (1, 0.01), (2, 0.02), (1, 0.04).
And the k+ reaches vector indicates that there are only 2 unique users that were reached by publisher1 -- user_1 and user_2.

src/simulator/halo_simulator.py
riemanli: Thanks for the suggestions! I like the second approach and will change the code to that.
Just note that I think the first approach is not correct because it will generate a list with only non-zero values. For example, 
```python
pub_subset = set([3, 5, 8])
impressions = list(range(10))
print([imps for pub_id, imps in enumerate(impressions) if pub_id in pub_subset])
>>> [3, 5, 8]
```
Instead of `[0 0 0 3 0 5 0 0 8 0]` which is the desired output.

src/simulator/tests/halo_simulator_test.py
riemanli: Fixed.

src/simulator/tests/halo_simulator_test.py
riemanli: Fixed.

src/simulator/halo_simulator.py
riemanli: Based on the discussion with @matthewclegg , we will change the code to the second approach which is suggested above.

riemanli: Done.

src/simulator/tests/halo_simulator_test.py
matthewclegg: You're right, thanks for correcting me on this.

matthewclegg (approved):

jiayu-google (approved):
Great job, Rieman!

RonPinelliMRC (approved):

PR #54, opened 2021-07-22
Sample reaches in the Venn diagram

riemanli:


src/simulator/halo_simulator.py
```diff
@@ -267,6 +267,43 @@ def simulated_venn_diagram_reach_by_spend(
...
+            for region_repr, kplus_reaches in primitive_regions.items()
+        ]
+        region_repr_seq, reach_population = list(zip(*region_repr_and_reach_pairs))
+
+        sampled_reach = random_generator.multivariate_hypergeometric(
+            reach_population, sample_size
+        )
+
+        return {
+            region_repr: [r] for region_repr, r in zip(region_repr_seq, sampled_reach)
```
riemanli, line 304: I am thinking to make the value in the dictionary just reach instead of 1+ reaches (ie. [reach]) when all the Venn diagram changes are gathered.

matthewclegg: +1

matthewclegg (changes requested):
I agree with your proposed change.

src/simulator/halo_simulator.py
jiayu-google: +1

src/simulator/tests/halo_simulator_test.py
```diff
@@ -113,6 +127,34 @@ def test_num_active_registers(self, cardinality, random_state, num_active):
...
+            "sample_size": 1,
+            "random_generator": np.random.default_rng(0),
+            "expected": {3: [1]},
+        },
+        {
+            "testcase_name": "with_5_regions_and_fake_rng",
+            "regions": {i: [i ** 2 + 1] for i in range(5)},
+            "sample_size": 20,
+            "random_generator": FakeRandomGenerator(),
+            "expected": {i: [n] for i, n in enumerate([1, 2, 5, 6, 6])},
```
jiayu-google, line 143: Why 1,2,5, 6, 6? 
IIU FakeRandomGenerator correctly, it'll be [0^2 + 1, 1^2 + 1, 2^2 + 1, 3^2 + 1, remaining] = [1, 2, 5, 10, 3]. Could you walk me through this?

jiayu-google (changes requested):
Thank you Rieman!

src/simulator/tests/halo_simulator_test.py
riemanli:
```python
class FakeRandomGenerator:
    def multivariate_hypergeometric(self, colors, nsample):
        samples = [0] * len(colors)
        index = 0

        while nsample:
            if samples[index] < colors[index]:
                samples[index] += 1
                nsample -= 1
            index = (index + 1) % len(samples)

        return samples
```

How the FakeRandomGenerator work is, it keeps going around the population list and draws one sample if the current slot has any items remained. 

Take the test case you quoted for example, the initial states are `colors = [1, 2, 5, 10, 17]`, `samples = [0, 0, 0, 0, 0]`, and `index = 0`. Inside the while-loop, `samples` changes as following:
    `[0, 0, 0, 0, 0]` -> `[1, 0, 0, 0, 0] (index = 0)` -> `[1, 1, 0, 0, 0] (index = 1)` -> `[1, 1, 1, 0, 0] (index = 2)` -> `[1, 1, 1, 1, 0] (index = 3)` -> `[1, 1, 1, 1, 1] (index = 4)`. Then `index` becomes 0 again. Since `samples[0]` is equal to `colors[0]`, i.e. no items available in slot-0, we do nothing and increment `index` to 1. Then,
    -> `[1, 2, 1, 1, 1]` -> ...  -> `[1, 2, 2, 2, 2]` -> `[1, 2, 3, 2, 2]` -> ...  -> `[1, 2, 5, 5, 5]` -> ... -> `[1, 2, 5, 6, 6]`. At this moment `nsample == 0`, exit the while-loop and return `samples`.

Hope that helps.

src/simulator/halo_simulator.py
riemanli: Thanks for the feedback, @jiayu-google and @matthewclegg . 

Since other functions like `_form_venn_diagram_regions`, `_generate_reach_points_from_venn_diagram`, and etc. also assumes 
```python 
primitive_regions: Dict[int, List]
```
Should I keep this one as what it is now, and then make the changes in another PR for all related functions?

jiayu-google: Good question. I don't think we need to change other places. 

The k+ reach of a primitive region or a subset of pubs is meaningful --- later we do want k+ reach to generate training points for frequency model. 

As for the `_sample_venn_diagram` method here, ideally we also want to include k+ reach; but, it is **infeasible**. (Precisely, the complete `primitive_regions: Dict[int, List]` does not contain enough info to obtain the sampled `Dict[int, List]`. To obtain the latter, we need info all the down from the raw DataSet.)

riemanli: I agree that k+ reaches will be useful in the future when we are ready. However, for now if we want to change the output type of `_sample_venn_diagram` to `Dict[int, int]`, the downstream functions need to change accordingly. For example, `_generate_reach_points_from_venn_diagram` is currently accepting `primitive_regions: Dict[int, **List**]`. If its input, which is the output from `_sample_venn_diagram`, changes to `Dict[int, **int**]`, we need to modify the code in `_generate_reach_points_from_venn_diagram` to accommodate the change.

My thought is:
`_form_venn_diagram_regions` still outputs regions with k+ reaches, so the functionality of obtaining k+ reaches still remains. Then let `_sampled_venn_diagram` output regions with sampled **reach**, and all the downstream functions use the regions with reach.

WDYT?

src/simulator/tests/halo_simulator_test.py
jiayu-google: I see I see. Sorry I mis-read the codes!
Thank you so much for the clear explanation! Let's leave it here for other reviewers.

src/simulator/halo_simulator.py
jiayu-google: Sounds like a good plan!

src/simulator/tests/halo_simulator_test.py
riemanli: No problem at all! Glad to help 😃

src/simulator/halo_simulator.py
riemanli: Offline discussion with @matthewclegg: Matthew suggested to keep the type of `primitive_regions` as `Dict[int, List]` for now. After all the functionalities of Venn diagram query are done, we can come back to change the code.

src/simulator/halo_simulator.py
```diff
@@ -267,6 +267,49 @@ def simulated_venn_diagram_reach_by_spend(
...
+            primitive_regions:  Contains k+ reaches in the regions. The k+
+              reaches for a given region is given as a list r[] where r[k] is
+              the number of people who were reached AT LEAST k+1 times.
+            sample_size:  The total number of sampled reach from the primitive
+              regions.
+            random_generator:  An instance of numpy.random.Generator that is
+              used for generating samples from a multivariate hypergeometric
+              distribution.
+        Returns:
+            A dictionary in which each key are the binary representations of
```
matthewclegg, line 288: You should probably note that the length of each list is 1.

matthewclegg (changes requested):
In looking at the documentation for numpy.random.Generator.multivariate_hypergeometric, I saw the following comment:

> The default method is “marginals”. For some cases (e.g. when colors contains relatively small integers), the “count” method can be significantly faster than the “marginals” method. If performance of the algorithm is important, test the two methods with typical inputs to decide which works best.

I was wondering if you had tried the two methods and assessed which would be better for our use case?

src/simulator/halo_simulator.py
riemanli: Added. In addition, I modified the description about `primitive_regions` in every function that has `primitive_regions` in its docstring. Note that `primitive_regions` before sampling contains k+ reaches in each region. After sampling, each region in `primitive_regions` has the reach in a list with length 1.

riemanli:
> In looking at the documentation for numpy.random.Generator.multivariate_hypergeometric, I saw the following comment:
>
> The default method is “marginals”. For some cases (e.g. when colors contains relatively small integers), the “count” method can be significantly faster than the “marginals” method. If performance of the algorithm is important, test the two methods with typical inputs to decide which works best.
>
> I was wondering if you had tried the two methods and assessed which would be better for our use case?

@matthewclegg, not sure how to reply under your comment quoted above, so I started another thread here. 

I am aware of that there are 2 generation algorithms, but I am not sure how to assess them. Do we have preference for performance or speed? Maybe @jiayu-google has a better idea which algorithm to use.

riemanli:
Based on the team's decision, the code of the Venn diagram reach estimate will be refactored. Convert the PR to draft status before the refactoring is done.

src/simulator/halo_simulator.py
riemanli: The type of `primitive_regions` will become `Dict[int, int]` during the refactoring.

src/simulator/halo_simulator.py
riemanli: After refactoring, the outputted regions from `_sample_venn_diagram` is `Dict[int, int]` where the key is the binary representation of a region and  the value is the reach in that region.

riemanli:
@jiayu-google @matthewclegg, this PR is ready for review now.

src/simulator/halo_simulator.py
```diff
@@ -307,6 +307,60 @@ def _form_venn_diagram_regions(
...
+        self,
+        primitive_regions: Dict[int, List],
+        sample_size: int,
+        random_generator: np.random.Generator = np.random.default_rng(),
+    ) -> Dict[int, int]:
+        """Return primitive regions with sampled reaches.
+
+        Args:
+            primitive_regions:  A dictionary in which each key is the binary
+              representations of a primitive region of the Venn diagram, and
```
matthewclegg, line 320: Delete 's' from 'representations'.

src/simulator/halo_simulator.py
```diff
@@ -307,6 +307,60 @@ def _form_venn_diagram_regions(
...
+              which belongs to pub_id-0 and id-2.
+              The k+ reaches for a given region is given as a list r[] where
+              r[k] is the number of people who were reached AT LEAST k+1 times.
+            sample_size:  The total number of sampled reach from the primitive
+              regions.
+            random_generator:  An instance of numpy.random.Generator that is
+              used for generating samples from a multivariate hypergeometric
+              distribution.
+        Returns:
+            A dictionary in which each key is the binary representations of
```
matthewclegg, line 335: Delete 's' from 'representations'.

matthewclegg (approved):

src/simulator/halo_simulator.py
riemanli: Thanks for catching that!

src/simulator/halo_simulator.py
riemanli: Done.

src/simulator/halo_simulator.py
riemanli: Done.

jiayu-google (approved):
LGTM, thanks Rieman!

RonPinelliMRC (approved):

PR #60, opened 2021-08-01
Add liquidlegions formulas.

jiayu-google:


src/simulator/halo_simulator.py
```diff
@@ -213,6 +214,46 @@ def simulated_reach_by_spend(
         kplus_reaches = ReachPoint.frequencies_to_kplus_reaches(frequencies)
         return ReachPoint(impressions, kplus_reaches, spends)
 
+    def _liquid_legions_cardinality_estimate_variance(self, n: int) -> float:
+        """Variance of cardinality estimate by unnoised LiquidLegions.
```
matthewclegg, line 218: Is there a reference that you can give for this formula?  Perhaps that would be helpful for people outside of our team.

src/simulator/halo_simulator.py
```diff
@@ -213,6 +214,46 @@ def simulated_reach_by_spend(
...
+        a = self._params.liquid_legions.decay_rate
+        c = a * n / m / (1 - np.exp(-a))
+        variance = (
+            expi(-c) - expi(-c * np.exp(-a)) - expi(-2 * c) + expi(-2 * c * np.exp(-a))
+        )
+        variance *= a * n ** 2 / m / (np.exp(-c) - np.exp(-c * np.exp(-a))) ** 2
+        variance -= n
+        return variance
+
+    def _liquid_legions_num_active_regions(
```
matthewclegg, line 238: I wonder if the name _liquid_legions_sample_num_active_regions would be more accurate?

matthewclegg (changes requested):

src/simulator/halo_simulator.py
jiayu-google: It is included in the PoPETs paper which has been accepted and will hopefully be published soon.  No public reference off the top of my head.  Maybe we can add a doc to the WFA repo (let me first confirm if the paper is formally accepted).

src/simulator/halo_simulator.py
jiayu-google: I believe that num_active_registers is the most commonly used name within Google, and sample_num_active_regions or simply `sampled_regions` also make sense to me.  Now I prefer `num_active_registers`, but is open to other options.

src/simulator/halo_simulator.py
```diff
@@ -213,6 +214,46 @@ def simulated_reach_by_spend(
...
+        self, n: int, random_generator: np.random.Generator = np.random.default_rng()
```
riemanli, line 239: I know we have done this several times, but I always wonder why we need to have a different random generator when there is one in `self._param`.

src/simulator/halo_simulator.py
jiayu-google: Added a TODO here.

src/simulator/halo_simulator.py
jiayu-google: Thank you for pointing this out!  Revised following your practice in PR 56.

riemanli (approved):

matthewclegg (approved):

src/simulator/halo_simulator.py
```diff
@@ -213,6 +214,50 @@ def simulated_reach_by_spend(
...
+          An observation of the number of active registers in the LiquidLegions.
+        """
+        m = self._params.liquid_legions.sketch_size
+        a = self._params.liquid_legions.decay_rate
+        register_probs = np.exp(-a * np.arange(m) / m)
+        register_probs /= sum(register_probs)
+        random_generator = np.random.default_rng(
+            seed=self._params.generator.integers(low=0, high=1e9)
+        )
+        return sum(random_generator.multinomial(n, register_probs) == 1)
```
riemanli, line 259: Would it be simpler to directly use `self._params.generator` for `multinomial`?

```diff
- random_generator = np.random.default_rng(
-     seed=self._params.generator.integers(low=0, high=1e9)
- )
- return sum(random_generator.multinomial(n, register_probs) == 1)
+ return sum(self._params.generator.multinomial(n, register_probs) == 1)
```

RonPinelliMRC:
Do @jiayu-google's most recent changes need internal review before MRC s/o?

riemanli (approved):

riemanli:
@RonPinelliMRC, thanks for raising the concern. 
Yes, they do. The internal review is now done and ready for MRC s/o. 
Thank you!

RonPinelliMRC (approved):

PR #62, opened 2021-08-02
Updates to Gamma Poisson model

matthewclegg:
1. Corrects the parameterization of the Gamma-Poisson distribution (thanks, Rieman)
2. Shifts the distribution so that it starts from 1.
3. Changes the optimization routines so that they are more likely to find the global minimum.
4.

### DIFF
--- a/src/models/gamma_poisson_model.py
+++ b/src/models/gamma_poisson_model.py
@@ -227,7 +227,8 @@ class GammaPoissonModel(ReachCurve):
         Returns:
           Expected size of impression inventory for a population of N users.
         """
-        return N * (1.0 + alpha * beta)
+        # N * (1.0 + alpha * beta)
+        return N * (1 + scipy.stats.nbinom.mean(alpha, 1.0 / (1.0 + beta)))
 
     def _expected_histogram(self, I, Imax, N, alpha, beta, max_freq=10):
         """Computes an estimated histogram.
--- a/src/models/gamma_poisson_model.py
+++ b/src/models/gamma_poisson_model.py
@@ -400,6 +400,11 @@ class GammaPoissonModel(ReachCurve):
         # starting points.
         Nvalues = Nmin / (np.arange(1, nstarting_points + 1) / (nstarting_points + 1))
 
+        # If the objective function for an optimization attempt is smaller than
+        # this value, then we have found a model that fits the data.  There is no
+        # need to continue searching for a better fit.
+        early_stopping_value = scipy.stats.chi2.ppf(0.98, len(h))
+
         for N0 in Nvalues:
             # Choose a reasonable starting point for optimization.
             Imax0, alpha0, beta0 = self._fit_histogram_fixed_N(
@@ -423,6 +428,9 @@ class GammaPoissonModel(ReachCurve):
             if result.fun < best_score:
                 best_alpha, best_beta, best_N = result.x
                 best_score = result.fun
+
+            if best_score < early_stopping_value:
+                break
 
         best_Imax = self._expected_impressions(best_N, best_alpha, best_beta)
         return best_Imax, best_N, best_alpha, best_beta
--- a/src/models/gamma_poisson_model.py
+++ b/src/models/gamma_poisson_model.py
@@ -143,7 +143,7 @@ class GammaPoissonModel(ReachCurve):
         This implementation makes use of the equivalence between the
         Gamma-Poisson distribution with parameters (alpha, beta) and
         the negative binomial distribution with parameters (p, r) =
-        (beta / (1 + beta), alpha).
+        (1 / (1 + beta), alpha).
 
         Args:
           n:  Value(s) at which the distribution is to be evaluated.
@@ -342,7 +342,7 @@ class GammaPoissonModel(ReachCurve):
 
         actual_oneplus = np.sum(h)
         predicted_oneplus = np.sum(hbar)
-        oneplus_error = (actual_oneplus - predicted_oneplus) ** 2
+        oneplus_error = one_plus_reach_weight * (actual_oneplus - predicted_oneplus) ** 2
 
         obj = np.sum((hbar - h) ** 2 / (hbar + 1e-6)) + oneplus_error
 
--- a/src/models/gamma_poisson_model.py
+++ b/src/models/gamma_poisson_model.py
@@ -342,7 +342,9 @@ class GammaPoissonModel(ReachCurve):
 
         actual_oneplus = np.sum(h)
         predicted_oneplus = np.sum(hbar)
-        oneplus_error = one_plus_reach_weight * (actual_oneplus - predicted_oneplus) ** 2
+        oneplus_error = (
+            one_plus_reach_weight * (actual_oneplus - predicted_oneplus) ** 2
+        )
 
         obj = np.sum((hbar - h) ** 2 / (hbar + 1e-6)) + oneplus_error
 
--- a/src/models/gamma_poisson_model.py
+++ b/src/models/gamma_poisson_model.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Gamma-Poisson reach curve model for underreported counts."""
 
+from absl import logging
 import numpy as np
 import scipy.stats
 from typing import List
@@ -22,11 +23,6 @@ from wfa_planning_evaluation_framework.models.reach_curve import ReachCurve
 # The following constant defines the highest frequency that will be
 # used for parameter estimation.
 MAXIMUM_COMPUTATIONAL_FREQUENCY = 1000
-
-# The ratio of the number of impressions observed in the data to the
-# size of the total impression inventory.  This ratio is kept constant
-# as a way of eliminating a free parameter in the model.
-IMPRESSION_INVENTORY_RATIO = 0.1
 
 
 class GammaPoissonModel(ReachCurve):
@@ -128,6 +124,12 @@ class GammaPoissonModel(ReachCurve):
         """
         if len(data) != 1:
             raise ValueError("Exactly one ReachPoint must be specified")
+        if data[0].impressions[0] < 0.001:
+            raise ValueError("Attempt to create model with 0 impressions")
+        if data[0].impressions[0] < data[0].reach(1):
+            raise ValueError(
+                "Cannot have a model with fewer impressions than reached people"
+            )
         self._reach_point = data[0]
         self._max_reach = max_reach
         if data[0].spends:
@@ -136,7 +138,7 @@ class GammaPoissonModel(ReachCurve):
             self._cpi = None
 
     def _logpmf(self, n, alpha, beta):
-        """Log of the PMF of the Gamma-Poisson distribution.
+        """Log of the PMF of the shifted Gamma-Poisson distribution.
 
         This implementation makes use of the equivalence between the
         Gamma-Poisson distribution with parameters (alpha, beta) and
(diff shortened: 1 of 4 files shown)


src/models/gamma_poisson_model.py
```diff
@@ -136,7 +138,7 @@ def __init__(self, data: List[ReachPoint], max_reach=None):
             self._cpi = None
 
     def _logpmf(self, n, alpha, beta):
-        """Log of the PMF of the Gamma-Poisson distribution.
+        """Log of the PMF of the shifted Gamma-Poisson distribution.
 
         This implementation makes use of the equivalence between the
         Gamma-Poisson distribution with parameters (alpha, beta) and
```
jiayu-google, line 144: shifted Gamma Poisson

src/models/tests/gamma_poisson_model_test.py
```diff
@@ -44,131 +44,136 @@ def gamma_poisson_pmf(k, alpha, beta):
...
-        self.assertAlmostEqual(gpm._alpha, 1.12, delta=0.05)
-
-    def test_fit_fixed_N(self):
+        self.assertAlmostEqual(gpm._max_reach, 10000, delta=1)
+        self.assertAlmostEqual(gpm._alpha, 1.0, delta=0.01)
+
+    @patch(
+        "wfa_planning_evaluation_framework.models.gamma_poisson_model.GammaPoissonModel._fit_histogram_fixed_N"
+    )
+    def test_fit_histogram_fixed_N(self, mock_gamma_poisson_model):
```
jiayu-google, line 153: There is another method above with exactly the same name `test_fit_histogram_fixed_N`

src/models/tests/gamma_poisson_model_test.py
```diff
@@ -44,131 +44,136 @@ def gamma_poisson_pmf(k, alpha, beta):
...
+        h_actual = [3749, 1875, 937, 468, 234, 117, 58, 29, 14, 7]
+        kplus_actual = [7488, 3739, 1864, 927, 459, 225, 108, 50, 21, 7]
+        rp = ReachPoint([15000], kplus_actual)
+        gpm = GammaPoissonModel([rp])
+        Imax, alpha, beta = gpm._fit_histogram_fixed_N(rp, 10000)
+        h_predicted = gpm._expected_histogram(15000, Imax, 10000, alpha, beta)
         for i in range(len(h_actual)):
             self.assertTrue(
-                (h_actual[i] - h_predicted[i]) ** 2 < 1,
+                (h_actual[i] - h_predicted[i]) ** 2 / h_predicted[i] < 1,
```
jiayu-google, line 133: Now i = 0 (i.e., 1 reach), the discrepancy is 0.236.
For all i >= 1, (h_actual[i] - h_predicted[i]) ** 2 / h_predicted[i] < 0.1.

How is it compared to the performance of the previous optimizer?

jiayu-google (changes requested):
Great job, Matt!

src/models/gamma_poisson_model.py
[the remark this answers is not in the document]
matthewclegg: Thanks.

src/models/gamma_poisson_model.py
```diff
@@ -247,116 +249,317 @@ def _impression_count(self, h):
...
+
+        Args:
+          h: Actual frequency histogram.  h[i] is the number of people
+             reached exactly i+1 times, except for h[-1] which is the
+             number of people reached len(h) or more times.
+          I: Total number of impressions purchased.
+          N: Total number of unique users.
+          alpha:  Parameter of the Gamma-Poisson distribution.
+          beta:  Second parameter of the Gamma-Poisson distribution.
+          one_plus_reach_weight:  Weight to be given to the squared
```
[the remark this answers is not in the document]
matthewclegg: Thanks for catching that!

src/models/gamma_poisson_model.py
```diff
@@ -247,116 +249,317 @@ def _impression_count(self, h):
...
+        predicted_oneplus = np.sum(hbar)
+        oneplus_error = (actual_oneplus - predicted_oneplus) ** 2
+
+        obj = np.sum((hbar - h) ** 2 / (hbar + 1e-6)) + oneplus_error
+
+        if np.isnan(obj):
+            logging.vlog(2, f"alpha {alpha} beta {beta} N {N} Imax {Imax}")
+            logging.vlog(2, f"h    {h}")
+            logging.vlog(2, f"hbar {hbar}")
+            raise RuntimeError("Invalid value of objective function")
```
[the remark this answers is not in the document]
matthewclegg: The list of built-in Python exceptions can be found here:  https://docs.python.org/3/library/exceptions.html

In theory, the objective function should never produce NaN values.  If it does, then this indicates some sort of algorithmic problem.  I like RuntimeError because it is a generic error that captures this circumstance.

I suppose some of the other candidates that seem like they could be close matches are as follows:  FloatingPointError, OverflowError, ArithmeticError.  We don't really know what the cause of the error is, so FloatingPointError and OverflowError seem like they might not be accurate.  ArithmeticError seems like a reasonable choice, though.  

Let me know if you would like me to change the exception type.

src/models/gamma_poisson_model.py
```diff
@@ -247,116 +249,317 @@ def _impression_count(self, h):
...
-        """Chi-squared fit to histogram h for fixed N.
+        # Search for local optimum on a grid whose initial spacing is
+        # given by grid_width
+        grid_width = 0.05
+
+        # Try to start from a point that is near the origin but also
+        # in the feasible region
+        alpha = beta = 0.1
+        if I0 > N:
+            alpha = beta = 1.1 * np.sqrt(I0 / N - 1) + grid_width
```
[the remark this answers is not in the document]
matthewclegg: The expected total number of impressions for N users is given by the formula:
   E[I] = N * (alpha * beta + 1)

In other words, E[I] / N - 1 = alpha * beta.  

In our case, I0 is the actual number of impressions that were shown to users.  This value has to be less than the total impression inventory, so I0 < E[I].  Thus, we have:

alpha * beta >= I0/N - 1.

Setting alpha = beta, we have alpha^2 >= I0/N - 1, or alpha >= np.sqrt(I0/N - 1).

The purpose of multiplying by 1.1 and adding grid_width was to ensure that the starting point is far enough away from this minimum value to avoid getting NaN values.  Probably we could delete the 1.1 factor, but the algorithm seems to work as it is.

src/models/gamma_poisson_model.py
```diff
@@ -247,116 +249,317 @@ def _impression_count(self, h):
...
+
+        score = 1e99
+        while grid_width > 0.01:
```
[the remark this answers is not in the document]
matthewclegg: Thanks!

src/models/tests/gamma_poisson_model_test.py
matthewclegg: Thanks for catching this!

src/models/tests/gamma_poisson_model_test.py
matthewclegg: Here are the parameter values:

```
True values:      Imax = 30000, alpha = 1, beta = 2
New optimizer: Imax = 30454, alpha = 1.03, beta = 1.98
Old optimizer:  Imax = 148880, alpha = 15.47, beta = 1.04

Here are the frequency histograms:
True: [3749, 1875, 937, 468, 234, 117, 58, 29, 14, 7]
New: [3719, 1886, 945, 471, 234, 116, 57, 28, 14, 7]
Old:  [3041, 2521, 1464, 669, 256, 85, 25, 6, 1, 0]
```

src/models/gamma_poisson_model.py
```diff
@@ -225,7 +227,7 @@ def _expected_impressions(self, N, alpha, beta):
         Returns:
           Expected size of impression inventory for a population of N users.
         """
-        return N * alpha / beta
+        return N * (1.0 + alpha * beta)
```
riemanli, line 230: Could you elaborate how the expected impression is equal to this line?

src/models/gamma_poisson_model.py
```diff
@@ -247,116 +249,319 @@ def _impression_count(self, h):
...
+            difference in 1+ reach values for the actual and estimated
+            histograms.
+        Returns:
+          chi^2 distance between the actual and expected histograms, plus
+          a term for weighting the difference in 1+ reach.
+        """
+        # Estimate total number of potential impressions
+        Imax = self._expected_impressions(N, alpha, beta)
+        if Imax <= I:
+            return np.sum(np.array(h) ** 2)
```
riemanli, line 332: Do you know in what scenarios we will have `Imax <= I`? 

In addition, I wonder the huge flat loss surface you mentioned in the slides is caused by these two lines which always outputs `np.sum(np.array(h) ** 2)`.

src/models/gamma_poisson_model.py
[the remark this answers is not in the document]
jiayu-google: Thanks for the illustration!   Now I also feel that RuntimeError is appropriate.

src/models/tests/gamma_poisson_model_test.py
jiayu-google: Impressive!  Leaving the comment open for other reviewers.

src/models/gamma_poisson_model.py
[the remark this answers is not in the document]
jiayu-google: This is cool!

jiayu-google (approved):

src/models/gamma_poisson_model.py
matthewclegg: I think I found the source of confusion.  According to Wikipedia, the PMF of the negative binomial is as follows:

f_wikipedia(k | p, r) = C(k + r - 1, k) (1 - p)^r p^k

However, scipy.stats.nbinom defines the PMF as follows:

f_scipy(k | p, r) = C(k + r - 1) p^r (1 - p)^k

In the two definitions, p has been swapped with 1-p.  In other words, f_scipy(k | p, r) = f_wikipedia(k | 1-p, r). 

From the Wikipedia page, the mean of a negative binomial with parameters p and r is pr / (1 - p).  However, replacing 1-p with p in this definition, the mean of a negative binomial according to scipy with parameters p and r is (1 - p) r / p.

When reparametrizing a Gamma-Poisson distribution as a negative binomial, we use the parameterization r = alpha, p = 1/(1 + beta).  In other words, beta = (1 - p) / p.  Thus, plugging these values into the formula for the mean, we get that the mean of the distribution is (1 - p) r / p = alpha * beta.

In our case, we add +1 because the distribution is shifted by 1.  And we multiply by N because N is the total number of users in the universe.

In any event, it doesn't really matter which parameterization we use, as long as everything is self-consistent.  I think the unit tests and the colab demonstrate that everything in this implementation is self-consistent.

I changed the code so that instead of using the explicit formula for the mean, it uses scipy.stats.nbinom.mean.  Hopefully this will remove all remaining doubt.

https://en.wikipedia.org/wiki/Negative_binomial_distribution
https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.nbinom.html

src/models/gamma_poisson_model.py
matthewclegg: I is the number of impressions that were actually served on behalf of the campaign.  Imax is the total number of impressions that could have been served if all impressions belonging to all users were purchased.  In principle, it should never be the case that Imax < I, because that would mean that more impressions were served than were available to  be served.  Since Imax = N * (1 + alpha * beta), it can be less than I if both alpha and beta are sufficiently small.  So, in the graphs of the loss surface, one sees a high wall along the borders of the surface, where alpha or beta or both are very small.  This high wall is the region where Imax < I.  The huge flat loss surface occurs in the interior, where I < Imax.

src/models/gamma_poisson_model.py
riemanli: Thank you so much for the detailed explanation! It all makes sense to me now.

src/models/gamma_poisson_model.py
riemanli: Got it. Thank you so much for explaining this to me.

riemanli (approved):
Great work! Thanks Matthew.

RonPinelliMRC (approved):

PR #64, opened 2021-08-04
Scale up reach in the Venn diagram regions

riemanli:


src/simulator/halo_simulator.py
```diff
@@ -440,6 +441,72 @@ def _add_dp_noise_to_primitive_regions(
...
+                budget.epsilon * privacy_budget_split,
+                budget.delta * privacy_budget_split,
+            ),
+            DP_NOISE_MECHANISM_DISCRETE_LAPLACE,
+            {"privacy_budget_split": privacy_budget_split},
+        )
+        self._privacy_tracker.append(noise_event)
+
+        # scaling factor = cardinality estimate / sum of reaches in the primitive regions
+        scaling_factor = cardinality_estimate / sum(primitive_regions.values())
```
jiayu-google, line 502: I feel it is helpful to raise an exception when `sum(primitive_regions.values()) <= 0`.

jiayu-google (commented):
LGTM!

src/simulator/halo_simulator.py
riemanli: That's a good point Jiayu. My thought is letting `simulated_venn_diagram_reach_by_spend` function to handle the scenario when the reaches in the primitive regions are all zeros. Therefore, the validity of the primitive regions is ensured. 
WDYT?

matthewclegg: Since this is a theoretically valid output of the algorithm, I would not be in favor of raising an exception.  In this case, the scaling_factor should be set to 0.  Here is an article on best practices for exception handling in python:

https://eli.thegreenplace.net/2008/08/21/robust-exception-handling/

matthewclegg (changes requested):

src/simulator/halo_simulator.py
jiayu-google: +1 to both of you.  "Setting scaling_factor to zero" SGTM.

riemanli: Will follow @matthewclegg's suggestion.

riemanli: Modified.

matthewclegg (approved):

jiayu-google (approved):
LGTM!

RonPinelliMRC (approved):

PR #66, opened 2021-08-04
Sets random seed for np.random.

matthewclegg:
This ensures repeatability across runs.  The PyDOE library (and perhaps others) use numpy.random, which requires the seed to be set with np.random.seed().


jiayu-google (approved):
Thanks a lot, Matt!

RonPinelliMRC (approved):

PR #70, opened 2021-08-10
Modifies Gamma-Poisson model to return valid reach points for high spends

matthewclegg:
Previously, if a spend value was given that exceeded the total cost of all impressions, the GammaPoisson model returned NaN's for all reach values.  This PR modifies the behavior of GammaPoisson model so that instead it returns the last point on the reach curve.

### DIFF
--- a/src/models/gamma_poisson_model.py
+++ b/src/models/gamma_poisson_model.py
@@ -583,7 +583,7 @@ class GammaPoissonModel(ReachCurve):
         if len(impressions) != 1:
             raise ValueError("Impressions vector must have a length of 1.")
         hist = self._expected_histogram(
-            impressions[0],
+            min(impressions[0], self._max_impressions - 1),
             self._max_impressions,
             self.max_reach,
             self._alpha,
--- a/src/models/tests/gamma_poisson_model_test.py
+++ b/src/models/tests/gamma_poisson_model_test.py
@@ -206,6 +206,18 @@ class GammaPoissonModelTest(absltest.TestCase):
                 f"Got {h_actual[i]} Expected {h_expected[i]}",
             )
 
+    @patch.object(
+        GammaPoissonModel, "_fit_histogram_fixed_N", return_value=(25000, 5.0, 2.0)
+    )
+    def test_overspend(self, mock_gamma_poisson_model):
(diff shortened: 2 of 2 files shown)


src/models/gamma_poisson_model.py
```diff
@@ -583,7 +583,7 @@ def by_impressions(
         if len(impressions) != 1:
             raise ValueError("Impressions vector must have a length of 1.")
         hist = self._expected_histogram(
-            impressions[0],
+            min(impressions[0], self._max_impressions - 1),
```
jiayu-google, line 586: I haven't found a particular reason for " - 1".  Could you explain?

jiayu-google (approved):
Thank you, Matt!

src/models/gamma_poisson_model.py
RonPinelliMRC: @matthewclegg appears to be an open question from @jiayu-google. Please resolve.

matthewclegg: Sorry, I should have explained this earlier.  This was added out of an abundance of caution.  If impressions[0] == self._max_impressions, then the probability that a person with n impressions is shown k of them on behalf of the campaign (see the function _knreach, line 158 and following) becomes:

C(n, k) 1**k 0**(n-k) f(n | alpha, beta),

where C(n, k) is the binomial coefficient counting the number of distinct subsets of size k that can be chosen from a set of size n, and f(n | alpha, beta) is the probability mass function of the Gamma-Poisson distribution with parameters alpha and beta.  In any event, the above expression contains the term 0**0 when n=k, and I was worried that this might cause a floating point exception.  To avoid having to worry about this issue, I clamped impressions[0] at self._max_impressions-1.  For realistic values of impressions[0] and self._max_impressions, the difference is infinitesimally small and the user would never notice it.  If you want, I can investigate further and see if this is really necessary.

jiayu-google: Sorry that I forgot to reply to this.  Thank you for the detailed explanation, and that's quite careful!  I feel the need to be cautious here, and let's go with what it is.

jiayu-google (approved):
Thanks Matt!

RonPinelliMRC (approved):

PR #71, opened 2021-08-11
Adds ground truth for reach curves as an evaluation option for M3

matthewclegg:
This PR adds the ability to compute the M3 model using the ground truth for the underlying reach curves.  This allows us to see how much modeling error is due to the single publisher models versus the restricted pairwise union reach curve model.


jiayu-google (approved):
Thank you so much Matt!

RonPinelliMRC (approved):

PR #72, opened 2021-08-12
Uses same seed to initialize each experimental trial.

matthewclegg:
The purpose of this change is to help ensure that different modeling strategies are compared using exactly the same inputs.

### DIFF
--- a/src/driver/experimental_trial.py
+++ b/src/driver/experimental_trial.py
@@ -90,7 +90,7 @@ class ExperimentalTrial:
         self._data_set_name = data_set_name
         self._trial_descriptor = trial_descriptor
 
-    def evaluate(self, rng: np.random.Generator) -> pd.DataFrame:
+    def evaluate(self, seed: int) -> pd.DataFrame:
         """Executes a trial.
 
         1. Check if the results for the trial have already been computed.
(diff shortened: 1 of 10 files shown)


src/driver/experimental_trial.py
```diff
@@ -114,6 +114,9 @@ def evaluate(self, rng: np.random.Generator) -> pd.DataFrame:
         logging.vlog(2, f"Dataset {self._data_set_name}")
         logging.vlog(2, f"Trial   {self._trial_descriptor}")
 
+        rng = np.random.default_rng(seed=seed)
+        np.random.seed(seed)
```
jiayu-google, line 118: Could you explain the function of this line in addition to the above line?

jiayu-google (approved):
Thanks for the hard work, Matt!

src/driver/experimental_trial.py
matthewclegg: In earlier distributions of numpy, random numbers were generated via RandomState.  This was subsequently replaced with Random Generator.  The Planning Evaluation Framework uses the more modern Random Generator throughout.  The seed for this generator is initialized at line 117.  However, some of the code that we depend on (specifically, PyDOE.lhs) uses the legacy RandomState.  The seed for this generator is initialized at line 118.

For more information, see the following:
https://numpy.org/doc/stable/reference/random/generator.html

RonPinelliMRC (approved):

PR #73, opened 2021-08-13
Modifies simulated_reach_by_spend so that it always generates consistent output

matthewclegg:
A basic property of k+ reach is that the k+ reach at frequency f is always nonnegative and at least as large as the k+ reach at frequency f+1.  Unfortunately, the StandardizedHistogramEstimator in the cardinality evaluation framework did not guarantee this property.  This PR modifies the output of StandardizedHistogramEstimator if necessary to ensure that the vector of k+ reach values is consistent.


jiayu-google (approved):
Wonderful numpy trick!

RonPinelliMRC (approved):

PR #77, opened 2021-08-26
Impression minimums for Gamma-Poisson distribution

matthewclegg:
When fitting noisy data, the Gamma-Poisson distribution tends to underestimate the true audience size.  This PR adds a penalty term to the objective function that penalizes parameter settings where the impression count is less than a specified factor times the observed impression count.

### DIFF
--- a/src/models/gamma_poisson_model.py
+++ b/src/models/gamma_poisson_model.py
@@ -334,6 +334,9 @@ class GammaPoissonModel(ReachCurve):
           chi^2 distance between the actual and expected histograms, plus
           a term for weighting the difference in 1+ reach.
         """
+        if alpha <= 0 or beta <= 0 or N <= 0:
+            return np.sum(np.array(h) ** 2)
+
         # Estimate total number of potential impressions
         Imax = self._expected_impressions(N, alpha, beta)
         if Imax <= I:
--- a/src/models/gamma_poisson_model.py
+++ b/src/models/gamma_poisson_model.py
@@ -125,7 +125,11 @@ class GammaPoissonModel(ReachCurve):
             be achieved.
           extrapolation_multiplier:  Int.  If specified, then a penalty term is
             introduced that penalizes models where the expected number of impressions
-            is less than extrapolation_multiplier * data[0].impressions[0].
+            in the inventory is less than extrapolation_multiplier times the
+            observed number of impressions.  Thus, if the extrapolation_multiplier
+            is 2, then optimizer prefers models where the number of impressions in
+            the inventory is at least twice the number of impressions that were
+            observed in the campaign.
         """
         if len(data) != 1:
             raise ValueError("Exactly one ReachPoint must be specified")
--- a/src/models/gamma_poisson_model.py
+++ b/src/models/gamma_poisson_model.py
@@ -113,7 +113,9 @@ class GammaPoissonModel(ReachCurve):
     found.
     """
 
-    def __init__(self, data: List[ReachPoint], max_reach=None):
+    def __init__(
+        self, data: List[ReachPoint], max_reach=None, extrapolation_multiplier=1.0
+    ):
         """Constructs a Gamma-Poisson model of underreported count data.
 
         Args:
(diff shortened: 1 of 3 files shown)


src/models/gamma_poisson_model.py
```diff
@@ -113,14 +113,19 @@ class GammaPoissonModel(ReachCurve):
...
+    ):
         """Constructs a Gamma-Poisson model of underreported count data.
 
         Args:
           data:  A list of consisting of a single ReachPoint to which the model is
             to be fit.
           max_reach:  Optional.  If specified, the maximum possible reach that can
             be achieved.
+          extrapolation_multiplier:  Int.  If specified, then a penalty term is
+            introduced that penalizes models where the expected number of impressions
```
jiayu-google, line 127: expected number of impressions in the inventory is less than the multiplier times the observed number number impressions, i.e., ....

src/models/tests/gamma_poisson_model_test.py
```diff
@@ -218,6 +218,19 @@ def test_overspend(self, mock_gamma_poisson_model):
         self.assertAlmostEqual(gpm.by_impressions([30000]).reach(1), 10000.0, delta=0.1)
         self.assertAlmostEqual(gpm.by_spend([300]).reach(1), 10000.0, delta=0.1)
 
+    def test_fit_histogram_chi2_distance_with_extrapolation(self):
+        # The following point corresponds to
```
jiayu-google, line 222: Perhaps better to say:  The following point is generated under ...

src/models/tests/gamma_poisson_model_test.py
```diff
@@ -218,6 +218,19 @@ def test_overspend(self, mock_gamma_poisson_model):
...
+        #  I, Imax, N, alpha0, beta0 = 5000, 30000, 10000, 1.0, 2.0
+        # This represents a distribution with mean = 3 and variance = 6.
+        kplus_actual = [3745, 933, 230, 0, 0, 0, 0, 0, 0, 0]
+        rp = ReachPoint([5000], kplus_actual)
+        gpm = GammaPoissonModel([rp])
+        Imax, _, _, _ = gpm._fit_histogram_chi2_distance(rp, nstarting_points=1)
+        gpm2 = GammaPoissonModel([rp], extrapolation_multiplier=4.0)
+        Imax2, _, _, _ = gpm2._fit_histogram_chi2_distance(rp, nstarting_points=1)
+        self.assertAlmostEqual(Imax, 16500, delta=100)
+        self.assertAlmostEqual(Imax2, 20000, delta=100)
```
jiayu-google, line 232: 1. Wondering what are the source(s) of this delta.  Is `scipy.optimize.minimize` the only source?
2. Does larger extrapolation_multiplier (say, 10) pull Imax closer to 30000?

jiayu-google (commented):
Thanks Matt, left some comments and questions.  Feel free to sync with me offline.

src/models/gamma_poisson_model.py
matthewclegg: I updated the doc.  Thanks for the suggestion!

src/models/tests/gamma_poisson_model_test.py
matthewclegg: I updated this as well.

src/models/gamma_poisson_model.py
jiayu-google: Awesome update. Thanks!

jiayu-google (approved):

src/models/tests/gamma_poisson_model_test.py
matthewclegg: The comment that I gave at the top of this function was not complete.  The input histogram was formed by starting with the histogram of expected values per the model parameters.  Then, I zeroed out all of the values starting with frequency 4.  This simulates the effect of adding differentially private noise to the frequency buckets.  The resulting value for Imax was then 16540, and for Imax2 was 19999.99.  If also tried a run with the extrapolation_multiplier set to 10.  In this case, Imax was returned as 50023.20.  The fact that it does not exactly match the target of 50000 could be due to a couple possible causes: (1) At this value, the relative change in the value of the objective function may be below the tolerance threshold used by the optimizer; or, (2) this represents a local minimum of the objective function.

jiayu-google: Thanks for the super clear explanations!

RonPinelliMRC (approved):

PR #78, opened 2021-08-27
K-inflated Gamma-Poisson distribution and model

matthewclegg:


src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,551 @@
...
+    We assume that each a_i >= 0, and that a_1 + a_2 + ... + a_k < 1.
+
+    Let GP(n | alpha, beta) be the PMF of the shifted Gamma-Poisson distribution
+    with parameters alpha, beta.  Also, define s = a_1 + a_2 + ... + a_k,
+    and define t = \sum_{i=1}^k GP(i | alpha, beta).  Then, the PMF of the
+    k-inflated Gamma-Poisson distribution is given as
+
+     kGP(n | alpha, beta, a_1, ..., a_k)
+          = a_n if 0 < n <= k,
+            (1-s) / (1-t) GP(n | alpha, beta) if n > k.
```
jiayu-google, line 45: Very smart model!

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,551 @@
...
+    be the estimated number of people that would be reached i times,
+    given alpha, beta and N.  The objective function that we
+    compute is as follows:
+
+      chi2(h, hbar) = \sum_i (h[i] - hbar[i])^2 / hbar[i]
+
+    One question is what value of k should be chosen.  We propose
+    to consider successive values of k starting from 0, and to stop
+    when a fit is found such that the value of the objective function
+    is smaller than the value of the chi^2 statistic with f_max degrees
```
jiayu-google, line 194: critical value, i.e., <to fill in>% quantile

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,551 @@
...
+
+        n = len(a)
+        s = np.sum(a)
+        t = np.sum(self._gamma_poisson_pmf(np.arange(1, n + 1), alpha, beta))
+        u = self._gamma_poisson_pmf(
+            np.arange(n + 1, MAXIMUM_COMPUTATIONAL_FREQUENCY), alpha, beta
+        )
+        pmf_tail = ((1 - s) / (1 - t)) * u
+        # self._pmf[n] = kGP(n | alpha, beta, a_1, ..., a_k)
+        self._pmf = np.array([0.0] + list(a) + list(pmf_tail))
```
jiayu-google, line 65: Why including 0?  This was not mentioned in the model description.

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,551 @@
...
+    with parameters alpha, beta.  Also, define s = a_1 + a_2 + ... + a_k,
+    and define t = \sum_{i=1}^k GP(i | alpha, beta).  Then, the PMF of the
+    k-inflated Gamma-Poisson distribution is given as
+
+     kGP(n | alpha, beta, a_1, ..., a_k)
+          = a_n if 0 < n <= k,
+            (1-s) / (1-t) GP(n | alpha, beta) if n > k.
+    """
+
+    def __init__(self, alpha, beta, a):
```
jiayu-google, line 48: type of variables

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,551 @@
...
+
+        Computes the function:
+
+            C(n, k) p^k (1 - p)^{n-k} kGP(n | alpha, beta, a).
+
+        where C(n, k) is the binomial coefficient n!/[k!(n-k)!].
+
+        Args:
+          k:  Number of impressions that were seen by the user.
+          n:  Available inventory for the user.
```
jiayu-google, line 111: inventory size

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,551 @@
...
+        the negative binomial distribution with parameters (p, r) =
+        (1 / (1 + beta), alpha).
+
+        Args:
+          n:  Value(s) at which the distribution is to be evaluated.
+            This can be a scalar or a numpy array.
+          alpha: Alpha parameter of the Gamma-Poisson distribution.
+          beta: Beta parameter of the Gamma-Poisson distribution.
+        Returns:
+          f(n | alpha, beta), where f is the Gamma-Poisson distribution.
```
jiayu-google, line 81: is the pmf of ...

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,551 @@
...
+
+        self._alpha = alpha
+        self._beta = beta
+        self._a = a
+
+        n = len(a)
+        s = np.sum(a)
+        t = np.sum(self._gamma_poisson_pmf(np.arange(1, n + 1), alpha, beta))
+        u = self._gamma_poisson_pmf(
+            np.arange(n + 1, MAXIMUM_COMPUTATIONAL_FREQUENCY), alpha, beta
```
jiayu-google, line 61: Clarify somewhere in the docstring that the model is truncated at MAXIMUM_COMPUTATIONAL_FREQUENCY

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,551 @@
...
+        """Probability that a random user receives k impressions.
+
+        Computes the function:
+            sum_{n=k}^{infinity} g(k, n | alpha, beta, I, I_max).
+        where g(k, n | alpha, beta, I, I_max) is the probability that
+        a randomly chosen user will have an available inventory of n
+        impressions, of which k are shown.
+
+        Args:
+          k:  np.array specifying number of impressions that were seen by the user.
```
jiayu-google, line 130: Smart to allow k to be array here and n to be array above!

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,551 @@
...
+    given alpha, beta and N.  The objective function that we
+    compute is as follows:
+
+      chi2(h, hbar) = \sum_i (h[i] - hbar[i])^2 / hbar[i]
+
+    One question is what value of k should be chosen.  We propose
+    to consider successive values of k starting from 0, and to stop
+    when a fit is found such that the value of the objective function
+    is smaller than the value of the chi^2 statistic with f_max degrees
+    of freedom, where f_max is the highest frequency that is recorded.
```
jiayu-google, line 195: Then, can k ever stop at or near fmax (or precisely, the min fmax across training points)?  In that case, the model would seem too flexible and have little to do with Gamma Poisson.

jiayu-google: Noticed that you set a kmax below.  I'd like to print out (k_opt, fmax, kmax) for comparison.  If you've investigated into it, would you please show me some examples?

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,551 @@
...
+    """
+
+    def __init__(
+        self, data: List[ReachPoint], kmax=10, debug=False, extrapolation_multiplier=1.0
+    ):
+        """Constructs a Gamma-Poisson model of underreported count data.
+
+        Args:
+          data:  A list of consisting of a single ReachPoint to which the model is
```
jiayu-google, line 204: check grammar

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,551 @@
...
+            to be fit.
+          kmax:  Maximum number of values of the PMF that are allowed to be set
+            arbitrarily.
+          debug:  If True, prints debug information as models are fit.
+          extrapolation_multiplier:  Float.  If specified, then a penalty term is
```
jiayu-google, line 209: In the following codes, isn't it only used to determine Imin when it's not specified?  Then I don't understand what it means by "penalizes models where the expected number of impressions is less than ..."

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,551 @@
...
+            introduced that penalizes models where the expected number of impressions
```
jiayu-google, line 210: expected number of impressions of the inventory

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Thanks!

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Thanks, updated.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Done.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Done.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Done.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Done.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Done.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Thanks!

src/models/kinflated_gamma_poisson_model.py
matthewclegg: I have done a number of ad hoc tests, but I don't have a table of values that I can share with you, although I could put one together if you like.  You are right that k can become arbitrarily large.  In fact, when I have tried it with data that is clearly not Gamma-Poisson, I have often observed that the algorithm terminates when k = f_max.  I would not say that the model is "too flexible".  Rather, I would say that the model becomes non-parametric.  In other words, when the data is close to a Gamma-Poisson distribution, then the algorithm will terminate with a low value of k.  But if the data is not close to a Gamma-Poisson, then the model acts like a non-parametric model.  So, in that sense, this model interpolates between these two types of models -- the parameter Gamma-Poisson vs. a fully non-parametric model.  I look forward to hearing your thoughts.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Done.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Done.

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,602 @@
...
+
+        The probability that a user is not reached, given that a fraction q
+        of available impressions is purchased, is given by
+           sum_{n=1}^{infinity} (1 - q)^n f(n),
+        where f(n) is the PMF of the above exponential poisson distribution.
+        This simplifies to (1 - p)(1 - q) / (1 - p(1 - q)).
+
+        The probability that a user is reached at least once is one minus
+        the above quantity.  Finally, the value of q is taken to be
+        I / ((beta + 1) * N), as the average number of impressions that a
```
jiayu-google, line 329: Would you raise error or something if q > 1?

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,602 @@
...
+
+    def _exponential_poisson_reach(self, I: float, N: float, beta: float) -> float:
+        """Estimates reach in an exponential poisson model.
+
+        Estimates reach when I impressions are purchased from a population
+        of N users where the impression count of each user is distributed
+        according to the exponential poisson distribution with parameter beta.
+
+        Under the exponential poisson model, the probability that a
+        user will have n impressions is (1 - p) * p^{n-1}, where p =
```
jiayu-google, line 316: is f(n) = ...
since you mentioned f(n) below

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,602 @@
...
+          I:  The hypothetical number of impressions purchased.
+          N:  The audience size.
+          beta: The parameter of the exponential-poisson distribution.
+        Returns:
+          The expected reach.
+        """
+        return N * I * (beta + 1) / (N + beta * (N + I))
+
+    def _exponential_poisson_beta(self, I: float, N: float, R: float) -> float:
+        """Estimate beta given impressions, audience size and reach."""
```
jiayu-google, line 342: Perhaps clarify that estimate beta by method of first moment, i.e., backsolving self._exponential_poisson_reach(I, N, beta) = R

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,602 @@
...
+        if I * R - N * (I - R) <= 0:
+            return 1e3
+        return (N * (I - R)) / (I * R - N * (I - R))
+
+    def _exponential_poisson_N_from_beta(
+        self, I: float, R: float, beta: float
+    ) -> float:
+        """Estimate audience size from impressions, reach and beta."""
+        if I == R:
```
jiayu-google, line 351: Change to I <= R or I < R + 1e-3 for safety?

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,602 @@
...
+
+        We know that the mean mu of the distribution has to be greater than
+        I / R.  Since beta = mu - 1, this says that beta >= I / R - 1.
+        But how much bigger?  That cannot easily be determined from just one
+        point.  Here, we assume beta = s * (I / R - 1), where s is some
+        small scale factor that is at least one.  From that, we estimate N.
+        """
+        if mu_scale_factor < 1.0:
+            raise ValueError("mu_scale_factor must be at least 1.0")
+        elif I == R:
```
jiayu-google, line 366: Consider <=  or < R + 1e-3

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,602 @@
...
+
+    def _exponential_poisson_N_from_beta(
+        self, I: float, R: float, beta: float
+    ) -> float:
+        """Estimate audience size from impressions, reach and beta."""
+        if I == R:
+            return 100.0 * R
+        return -(I * beta * R) / ((beta + 1) * (R - I))
+
+    def _exponential_poisson_N(self, I: float, R: float, mu_scale_factor=1.5) -> float:
```
jiayu-google, line 355: why using mu instead of directly using beta?

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Thanks for pointing this out.  I decided to simply return the maximum possible reach N in this case.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Good catch.  Thanks.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Done.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Good idea.  Done.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Good idea.  Done.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: In my view, beta is an internal model parameter that is hard to interpret, whereas mu is a quantity that is very easy to understand.  So, it seemed more natural to formulate this value in terms of mu.  I can change it to beta if you think that would be better.  Let me know.

src/models/kinflated_gamma_poisson_model.py
jiayu-google: LGTM!

src/models/kinflated_gamma_poisson_model.py
jiayu-google: Makes sense to use mu here.

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,633 @@
...
+          is an estimated KInflatedGammaPoissonDistribution representing an
+          exponential-poisson distribution.
+        """
+        impressions = point.impressions[0]
+        if Imin is not None and impressions < Imin:
+            impressions = Imin
+        reach = point.reach()
+        if len(point.frequencies) > 1:
+            # Estimate the mean from the distribution found in the reach point
+            mu = impressions / min(reach, impressions-1)
```
jiayu-google, line 430: perhaps `max(1, min(reach, impressions - 1))`

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,633 @@
...
+        Returns:
+          A pair (N, dist), where N is the estimated audience size and dist
+          is an estimated KInflatedGammaPoissonDistribution representing an
+          exponential-poisson distribution.
+        """
+        impressions = point.impressions[0]
+        if Imin is not None and impressions < Imin:
+            impressions = Imin
+        reach = point.reach()
+        if len(point.frequencies) > 1:
```
jiayu-google, line 428: I'm confused why we don't use `self._exponential_poisson_N` when `len(point.frequencies) > 1`.

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,633 @@
...
+        be a reasonable match at these as well, although the match is not
+        guaranteed to be perfect.
+
+        The parameters returned by this function are used to bootstrap the
+        full k-Inflated Gamma Poisson model.
+
+        Args:
+          point:  A ReachPoint to which an exponential-poisson model is
+            to be fit.
+          Imin:  Minimum number of impressions that the returned model
```
jiayu-google, line 417: Minimum expected number of inventory impressions

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,633 @@
...
+            should have.
+        Returns:
+          A pair (N, dist), where N is the estimated audience size and dist
+          is an estimated KInflatedGammaPoissonDistribution representing an
+          exponential-poisson distribution.
+        """
+        impressions = point.impressions[0]
+        if Imin is not None and impressions < Imin:
```
jiayu-google, line 425: Why forcing `impressions` to be `>= Imin` here, instead of forcing `Imax >= Imin` later?

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,633 @@
...
+        Returns:
+          chi^2 distance between the actual and expected histograms, plus
+          a term for weighting the difference in 1+ reach.
+        """
+        # Estimate total number of potential impressions
+        Imax = N * dist.expected_value()
+        if Imax <= I:
+            return np.sum(np.array(h) ** 2)
+
+        freqs = N * dist.kreach(np.arange(1, len(h)), I / Imax)
```
jiayu-google, line 297: We should have `I >= sum([(k + 1) * h[k] for k in range(len(h))`.  Correct me if I'm wrong.  
You may check if this condition is always satisfied.  If it is violated, will it introduce bug?

src/models/kinflated_gamma_poisson_model.py
```diff
@@ -0,0 +1,633 @@
...
+
+        if point.max_frequency == 1:
+            return N0, dist
+
+        N, dist, score = self._fit_histogram_fixed_length_a(
+            h, I, Imin, N0, dist._alpha, dist._beta, []
+        )
+        self.print_fit("start", score, N, dist._alpha, dist._beta, dist._a)
+
+        while scipy.stats.chi2.cdf(score, len(h)) > 0.95 and len(dist._a) < self._kmax:
```
jiayu-google, line 553: Is it necessary to set a max num_iterations?

jiayu-google, line 553: Is `scipy.stats.chi2.cdf(score, len(h))` always decreasing as len(a) increases?

src/models/kinflated_gamma_poisson_model.py
matthewclegg: If Imax <= I, then there will be an error in the computation of the histogram, because the probability that a given impression is selected is I / Imax, and this value would be greater than 1.  So, for that reason, it's important to check this boundary condition.  If I < sum([(k + 1) * h[k] for k in range(len(h)), then the result will be that the quality of the fit is poor and the value of the objective function is high.  It will not cause an error (to my knowledge).  So, in my view, this does not need to be checked separately.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Done.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: Imin could be None, so comparing Imax to Imin would not make sense in that case.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: If additional frequencies are given, then we try to estimate the average frequency per user, and use this to initialize beta.  If only frequency 1 is given, then we could be in one of two cases: (1) every user was reached exactly once (seems unlikely), or (2) due to the addition of differentially private noise, we don't know how many people were reached at frequencies 2 and higher.  In the latter case, we make an arbitrary starting assumption about the average frequency.  This is represented by the mu_scale_factor parameter in _exponential_poisson_N.  If the assumption turns out to be incompatible with the data, it is revised upwards in the while loop below.  So, in my view, it is better to use self._exponential_poisson_N_from_beta when there is data to support it, because it makes fewer arbitrary assumptions.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: The denominator of this expression is always less than `impressions`, so the expression is guaranteed to have a value that is at least 1.

src/models/kinflated_gamma_poisson_model.py
matthewclegg: In this specific algorithm, the answer to that question is yes.  The reason is that each new iteration is initialized from the best values that were found in the previous iteration.  In other words, when a is extended in line 554, the new value that is added to a is the value of the Gamma-Poisson PMF at that point.  The optimizer is guaranteed to return a point where the objective function is no greater than the objective function at the starting point.  Consequently, the values of `score` are guaranteed to be decreasing.  Since `scipy.stats.chi2.cdf` is a monotonic function, it follows that `scipy.stats.chi2.cdf(score, len(h))` is guaranteed to decrease with each iteration.

src/models/kinflated_gamma_poisson_model.py
jiayu-google: Got it, sorry I misunderstood.

src/models/kinflated_gamma_poisson_model.py
jiayu-google: Thanks for detailed explanation, Matt!

src/models/kinflated_gamma_poisson_model.py
jiayu-google: Got it!

jiayu-google (approved):
Approved after reviewing this PR and also the results (that Matt presented to WFA).

RonPinelliMRC (approved):

PR #80, opened 2021-09-02
Removes various forms of non-determinism.

matthewclegg:
Also, adds caching to DataDesign.py which improves performance.
Also, adds code for keeping track of which trials are currently under evaluation, making it easier to debug.

### DIFF
--- a/src/driver/experimental_trial.py
+++ b/src/driver/experimental_trial.py
@@ -138,7 +138,8 @@ class ExperimentalTrial:
             return pd.read_csv(trial_results_path)
 
         # The pending directory contains one entry for each currently executing
-        # experimental trial.  Each d
+        # experimental trial.  If a computation appears to hang, this can be
+        # used to check which evaluations are still pending.
         experiment_dir_parent = Path(self._experiment_dir).parent
         pending_path = Path(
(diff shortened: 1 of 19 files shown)


src/driver/test_point_generator.py
```diff
@@ -37,6 +37,12 @@
 # give an acceptably small sampling variance for the modeling errors.
 MINIMUM_NUMBER_OF_TEST_POINTS = 100
```
jiayu-google, line 38: Would you remove this since you added MINIMUM_NUMBER_OF_TEST_POINTS_PER_PUBLISHER

src/driver/experimental_trial.py
```diff
@@ -135,6 +137,17 @@ def evaluate(self, seed: int) -> pd.DataFrame:
             logging.vlog(2, "  --> Returning previously computed result")
             return pd.read_csv(trial_results_path)
 
+        # The pending directory contains one entry for each currently executing
+        # experimental trial.  Each d
```
jiayu-google, line 141: Complete this sentence please

jiayu-google (approved):
Thanks, this has become much clearer.

src/driver/experimental_trial.py
matthewclegg: Done.  Thanks for catching this.

RonPinelliMRC (approved):

PR #82, opened 2021-09-09
Records the maximum frequency for both actual data and halo output.

matthewclegg:


src/models/gamma_poisson_model.py
```diff
@@ -131,6 +131,7 @@ def __init__(
             the inventory is at least twice the number of impressions that were
             observed in the campaign.
         """
+        super().__init__(data, max_reach)
```
jiayu-google, line 134: This line is Ok but not useful?

`self._data` seems not used anywhere.   `self._max_reach` has been specified in line #144 below.

src/models/goerg_model.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, data: [ReachPoint]):
           max_reach:  Optional.  If specified, the maximum possible reach that can
             be achieved.
         """
+        super().__init__(data)
```
jiayu-google, line 40: Ditto

src/models/tests/reach_curve_test.py
```diff
@@ -44,20 +44,24 @@ def by_spend(self, spend: [float], max_frequency: int = 1) -> ReachPoint:
 class ReachCurveTest(absltest.TestCase):
     def test_reach(self):
         model = FakeReachCurve([ReachPoint([200], (100,))])
+        model._fit()
```
jiayu-google, line 47: Note that FakeReachCurve belongs to ReachCurves which belongs to ReachSurface.
`ReachSurface` has done `_fit()` in `__init__()`.  See line 42 of https://github.com/world-federation-of-advertisers/planning-evaluation-framework/blob/abe1c96ec93cf4d6c2d04808d8211c7c770739d2/src/models/reach_surface.py#L24

So this line is ok but not necessary?

jiayu-google (approved):
LGTM but left minor comments about possibly redundant codes.

src/models/gamma_poisson_model.py
matthewclegg: See `src/driver/experimental_trial.py`, line 300.

src/models/goerg_model.py
matthewclegg: See src/driver/experimental_trial.py, line 300.

src/models/tests/reach_curve_test.py
matthewclegg: Line 42 of `reach_surface.py` is deleted in this PR.  After this PR, it is necessary to always explicitly call `_fit()` after a model is instantiated.

RonPinelliMRC (approved):

PR #84, opened 2021-09-20
Include more metrics in the result dataframe

jiayu-google:


src/driver/experimental_trial.py
```diff
@@ -212,6 +212,35 @@ def evaluate(self, seed: int) -> pd.DataFrame:
             ],
             axis=1,
         )
+        if self._trial_descriptor.modeling_strategy == "m3strategy":
```
matthewclegg, line 215: Move this code to a function.  Also, is there something specific to the m3strategy in this code?  If not, remove the check for m3strategy.

src/driver/experimental_trial.py
```diff
@@ -212,6 +212,35 @@ def evaluate(self, seed: int) -> pd.DataFrame:
...
+            union_inventory = self._dataset.maximum_reach
+            true_single_inventories = [p.max_reach for p in self._dataset._data]
+            predicted_single_inventories = [
+                c._max_reach for c in self.reach_surface._reach_curves
+            ]
+            result["union_inventory_reach"] = union_inventory
+            result["true_single_pub_inventory_reaches"] = [true_single_inventories]
```
matthewclegg, line 222: A list of a list?  Can this be represented in a DataFrame?  Maybe it is best not to include this.

src/driver/experimental_trial.py
```diff
@@ -212,6 +212,35 @@ def evaluate(self, seed: int) -> pd.DataFrame:
...
+            result["predicted_single_pub_inventory_reaches"] = [
+                predicted_single_inventories
```
matthewclegg, line 224: Same comment as for true_single_pub_inventory_reaches.

matthewclegg (changes requested):
Please add a unit test.

Also, I would favor just adding a single new value, say 'overlap_ratio', which is the  the total union reach of all publishers divided by the sum of the maximum reaches of each publisher.

Perhaps you could also have a second statistic, 'predicted_overlap_ratio', which is the total total union of all publishers divided by the sum of  the predicted maximum reaches.

src/driver/experimental_trial.py
jiayu-google: I'm trying to save the list `true_single_inventories` into the df.   Directly letting `result["true_single_pub_inventory_reaches"] = true_single_inventories` causes bug, since pandas would try to put the list `true_single_inventories` into multiple rows.  So I did `[true_single_inventories]` to put them in a row.

If it is not desirable to assign a list to a df cell, here're alternative off the top of my head:
(1) convert the list into a tuple.
(2) convert the list into str like '10, 10, 10'.
It is desirable to keep all info of the list for analysis of prediction error.

src/driver/experimental_trial.py
jiayu-google: There's something specific to the m3strategy, which unfortunately makes the codes ugly.  Here I use `self.reach_surface._reach_curves`, which is an attribute in multi-pub models but not in other ReachSurfaces.  Specifically, not in single pub models nor in the `FakeSurface` in the unit test. 

Sure, will try to move this into a function and add unit test.

src/driver/experimental_trial.py
matthewclegg: My main concern is that I am not sure that we will be able to read and write this data to the CSV files that we are using for storing the intermediate DataFrames.  If this works, then I am fine with it as is.  Otherwise, you might need to convert it to a string list as you suggested.

PR #85, opened 2021-09-20
Modifies Halo simulator to return a consistent output when impressions <= reach

matthewclegg:


src/simulator/halo_simulator.py
```diff
@@ -210,6 +210,8 @@ def simulated_reach_by_spend(
 
         # convert result to a ReachPoint
         impressions = self._data_set.impressions_by_spend(spends)
+        if np.sum(impressions) <= kplus_reaches[0]:
+           kplus_reaches = [max(np.sum(impressions)-2, 1), 1] + [0] * max(0, max_frequency-2)
```
jiayu-google, line 214: 1. Why having 1 for the 2+ reach?
2. The indent = 3 in this line. Would it introduce any issue?

matthewclegg: 1. I think you are right, it would be simpler just to have a 1+ reach number.  I modified the code accordingly.
2. Thanks for pointing this out.

amiragha (approved):

PR #87, opened 2021-10-01
Add support for sampling buckets with basic composition

pasin30055:
Add support for sampling buckets in PrivacyTracker with basic composition. The code is the same as proposed design except that the buckets are represented as interval [0, 1) instead of integers (which helps e.g. avoid having to specify the total number of buckets at the beginning).

(Note that the code for privacy_consumption is not the most optimal since it is faster to do a single sweep over _sampling_buckets_starting_points and update the current budget consumption, but this approach does not directly work with PLD and anyway this part is unlikely to constitute a large computation resource in the framework. So I'm going with the simpler solution for the moment.)

### DIFF
--- a/src/simulator/privacy_tracker.py
+++ b/src/simulator/privacy_tracker.py
@@ -16,6 +16,7 @@
 from typing import Dict
 from typing import List
 from typing import NamedTuple
+from typing import Optional
 
 DP_NOISE_MECHANISM_LAPLACE = "Laplace"
 DP_NOISE_MECHANISM_DISCRETE_LAPLACE = "Discrete Laplace"
@@ -42,6 +43,27 @@ class PrivacyBudget(NamedTuple):
     delta: float
 
 
+class SamplingBucketIndices(NamedTuple):
+    """Records the indices of sampling buckets.
+
+    For simplicity, we represent buckets as an interval [0, 1). A user is
+    mapped to a real number in this interval. A sampling bucket represents an
+    interval [smallest_index, smallest_index + sampling_rate). This means that
+    all users that get mapped to this interval is used for computing
+    differentially private estimates.
+    """
+
+    smallest_index: float
+    sampling_rate: float
+
+    def contains(self, sampling_bucket: float) -> bool:
+        """Returns whether the sampling bucket is contained in the interval."""
+        return (
+            self.smallest_index <= sampling_bucket
+            and sampling_bucket < self.smallest_index + self.sampling_rate
+        )
+
+
 class NoisingEvent(NamedTuple):
     """Records the addition of differentially private noise
 
@@ -56,6 +78,8 @@ class NoisingEvent(NamedTuple):
     budget: PrivacyBudget  # Privacy budget associated to this event.
     mechanism: str  # See DP_NOISE_MECHANISM above.
     params: Dict  # Mechanism-specific parameters.
+    # See SamplingBucketIndices above.
+    sampling_buckets: SamplingBucketIndices = SamplingBucketIndices(0, 1)
 
 
 class PrivacyTracker:
@@ -77,9 +101,9 @@ class PrivacyTracker:
 
     def __init__(self):
         """Returns an object for recording and tracking privacy budget usage."""
-        self._epsilon_sum = 0.0
-        self._delta_sum = 0.0
         self._noising_events = []  # A list of NoisingEvents
+        # A list of starting points of buckets
+        self._sampling_buckets_starting_points = set([])
 
     @property
     def privacy_consumption(self) -> PrivacyBudget:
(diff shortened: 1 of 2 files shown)


src/simulator/privacy_tracker.py
```diff
@@ -89,7 +112,26 @@ def privacy_consumption(self) -> PrivacyBudget:
         basic composition rule.  This will be expanded in the future
         to support advanced composition (see TODO #1 above).
         """
-        return PrivacyBudget(self._epsilon_sum, self._delta_sum)
+        # Take maximum privacy budget spent across all buckets.
+        max_epsilon = 0
```
matthewclegg, line 116: I am a little conflicted about this.  I am not sure if `privacy_consumption` as it is defined will yield useful information.  My current thinking is that a more useful function would be as follows:

```
def remaining_query_count(epsilon):
  """Returns the number of additional queries, each having budget epsilon, that could be supported."""
```

What are your thoughts?

src/simulator/tests/privacy_tracker_test.py
```diff
@@ -66,6 +67,146 @@ def test_two_events(self):
             ],
         )
 
+    @parameterized.named_parameters(
```
matthewclegg, line 70: Nice use of parameterized testing.

matthewclegg (changes requested):
Sorry that this took so long!

src/simulator/privacy_tracker.py
```diff
@@ -89,7 +112,26 @@ def privacy_consumption(self) -> PrivacyBudget:
         basic composition rule.  This will be expanded in the future
         to support advanced composition (see TODO #1 above).
```
jiayu-google, line 113: Minor comment about previous PR...  where is TODO #1?  Will it make sense to move this comment to the `privacy_consumption_for_sampling_bucket` method below

src/simulator/privacy_tracker.py
```diff
@@ -42,6 +42,27 @@ class PrivacyBudget(NamedTuple):
     delta: float
 
 
+class SamplingBucketIndices(NamedTuple):
+    """Records the indices of sampling buckets.
+
+    For simplicity, we represent buckets as an interval [0, 1). A user is
+    mapped to a real number in this interval. A sampling bucket represents an
+    interval [smallest_index, smallest_index + sampling_rate). This means that
```
jiayu-google, line 50: As Matt mentioned below, wrap-arounds are possible.

I don't know what's the best way to clarify it here, perhaps just say [smallest_index, smallest_index + sampling_rate) mod [0, 1)...

src/simulator/privacy_tracker.py
```diff
@@ -89,7 +112,26 @@ def privacy_consumption(self) -> PrivacyBudget:
         basic composition rule.  This will be expanded in the future
         to support advanced composition (see TODO #1 above).
         """
-        return PrivacyBudget(self._epsilon_sum, self._delta_sum)
+        # Take maximum privacy budget spent across all buckets.
+        max_epsilon = 0
+        max_delta = 0
+        for sampling_bucket_ in self._sampling_buckets_starting_points:
```
jiayu-google, line 118: logically we shall take maximum over all x in [0, 1), but it seems enough to look at just starting points? 

This is smart! Would you consider proving or explaining it here.

src/simulator/privacy_tracker.py
```diff
@@ -42,6 +42,27 @@ class PrivacyBudget(NamedTuple):
...
+    mapped to a real number in this interval. A sampling bucket represents an
+    interval [smallest_index, smallest_index + sampling_rate). This means that
+    all users that get mapped to this interval is used for computing
+    differentially private estimates.
+    """
+
+    smallest_index: float
+    sampling_rate: float
+
+    def contains(self, sampling_bucket: float) -> bool:
```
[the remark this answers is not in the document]
pasin30055: Great catch! Added handling & test for wrap around.

src/simulator/privacy_tracker.py
pasin30055: TODO #1 is under PrivacyTracker. I've added the comment also in privacy_consumption_for_sampling_bucket so that we don't forget to remove there; but I'm also keep this comment here for clarity.

src/simulator/privacy_tracker.py
pasin30055: I think this is highly related to the usecase. I'd imagine both can be useful--depending on how the algorithm is interacting with the PrivacyTracker. In any case, we can definitely add remaining_query_count in a follow-up PR.

src/simulator/privacy_tracker.py
pasin30055: Added a comment to clarify this.

src/simulator/tests/privacy_tracker_test.py
pasin30055: Thanks!

src/simulator/privacy_tracker.py
```diff
@@ -42,6 +42,27 @@ class PrivacyBudget(NamedTuple):
     delta: float
 
 
+class SamplingBucketIndices(NamedTuple):
+    """Records the indices of sampling buckets.
+
+    For simplicity, we represent buckets as an interval [0, 1). A user is
+    mapped to a real number in this interval. A sampling bucket represents an
+    interval [smallest_index, smallest_index + sampling_rate). This means that
+    all users that get mapped to this interval is used for computing
```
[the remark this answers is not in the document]
pasin30055: Fixed. Thanks!

src/simulator/privacy_tracker.py
pasin30055: Changed as suggested now + added handling code & some test for it.

src/simulator/privacy_tracker.py
matthewclegg: Seems reasonable.  Let's go with this for now.

PR #90, opened 2021-10-02
Adds code for computing relative errors at higher frequencies for single pub models

matthewclegg:


src/driver/experimental_trial.py
```diff
@@ -295,6 +296,18 @@ def _compute_single_publisher_fractions_dataframe(
             column_name = f"relative_error_at_{int(r*100):03d}"
             results[column_name] = [relative_error]
 
+        for f in range(2, max_frequency):
```
jiayu-google, line 299: How about f = 1? It seems not to hurt by recording f=1.

jiayu-google (approved):
LGTM with a minor comment.

amiragha:
This looks good to me, having the same Q as Jiayu

src/driver/experimental_trial.py
matthewclegg: The case f=1 was already covered by the previous code, so there was no need to duplicate it.  Nonetheless, I have changed this to start from f=1.

jiayu-google (approved):
LGTM, thanks Matt!

amiragha (approved):

PR #96, opened 2021-11-09
Add simulation framework for pure sampling & sampling + LL

pasin30055:
- The file `noisy_liquid_legions_per_query_epsilon.py` can produce, via simulation, different type of errors given number of buckets involved in sampling, per-query epsilons and deltas (for DP), true reach, frequency distribution and other parameters of the liquid legions sketch. When the sketch size is set to -1, assume that no sketch is used.
- The file `per_query_vs_total_epsilon.py` can produce the maximum number of queries _per bucket_ for a given per-query epsilon, delta and total epsilon. 
- Directory `simulation_results` contain the simulated error results for `sketch_size=100_000` and `decay_rate=10` which are the parameters we propose to use for liquid legions.


jiayu-google (approved):
Sorry for getting back to you late, Pasin!

LGTM (based on my previous review on the same google internal codes and a quick check over this PR).

pasin30055:
@amiragha If you are available, would you mind taking a look at this PR? Please let me know if you have any questions. Thank you very much!

matthewclegg (approved):

PR #100, opened 2021-12-16
Analytical formulas for Sampling + LiquidLegions.

jiayu-google:
Just FYI, no pressure to review it.

Hi @pasin30055, just let you know that I haven't based this PR on your branch `pasin30055:sampling_research` since I don't know how to base on a branch in another repo...  If everything else works, I'll first check in the codes here and then add more tests based on your codes.  Hope it works!

### DIFF
--- a/src/accuracy/analytical_variance_for_sampling_liquid_legions.py
+++ b/src/accuracy/analytical_variance_for_sampling_liquid_legions.py
@@ -1,0 +1,186 @@
+# Copyright 2021 The Private Cardinality Estimation Framework Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Analytical variance formulas for sampling + LiquidLegions."""
+
+import numpy as np
+from numpy import exp as exp
+from scipy.special import expi as ei
+
+
+class AnalyticalAccuracyEvaluator:
+
+    """Analytical formulas to quickly evaluate the accuracy of sampling + LiquidLegions.
+
+    To estimate reach and frequency histogram using sampling + LiquidLegions, we
+        1. ask each publisher to sample a fraction of vids using the same hash.
+        2. ask each publisher to obtain the the LiquidLegions of the sampled, reached vids, with the same
+            decay rate and number of registers
+        3. Use MPC + SameKeyAggregator protocol to estimate reach and relative frequency histogram of the
+            sampled ids.  Scale up the reach estimate.  Use the relative frequency histogram of the sample
+            as the estimate for population.
+    This class computes the standard deviations of the estimates of
+        - reach, and
+        - proportion of each frequency bucket.
+    TODO(jiayu): derive the stds for k+ reaches and shuffle distance.
+
+    Please read Section of 1 the companion doc:
+    https://drive.google.com/file/d/1FNeMZnSYNMcsAMtzFY2ZyM_Tl1YYoEid/view?usp=sharing&resourcekey=0-NTSWpzbL2HkumqI-AUZlIg
+    for precise problem description and formulas.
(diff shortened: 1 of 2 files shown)


src/accuracy/analytical_variance_for_sampling_liquid_legions.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        self.a = a
+        self.m = int(m)
+        self.pi = pi
+        self.ssreach = ssreach
+        self.ssfreq = ssfreq
+        self.c = a / m / (1 - exp(-a))
+
+    def _local_multiply(self, n, multiplier):
+        """An operation repeatedly used in the formulas."""
```
pasin30055, line 60: This doesn't seem to be a very useful docstring. Is there a nice interpretation of this quantity? If so, then it might be worthwhile to spell that out instead. A similar comment applies to other _local_* methods below too.

src/accuracy/analytical_variance_for_sampling_liquid_legions.py
```diff
@@ -0,0 +1,186 @@
...
+    def _vy(self, n):
+        """Var(Y) in formula (11)."""
+        return n * self.pi * (1 - self.pi)
+
+    def _ea(self, ey):
+        """Expected number of active registers.  Formula (6)."""
+        return ey * self._theta_1(ey - 1)
+
+    def get_ea(self, n):
+        """Print expected number of active registers.  May be used as it is an important metric."""
```
pasin30055, line 129: "Print" -> "Returns"

pasin30055 (approved):
Thanks a lot Jiayu. This is very nice. Only a few minor comments below.

src/accuracy/analytical_variance_for_sampling_liquid_legions.py
matthewclegg: +1

src/accuracy/tests/analytical_variance_for_sampling_liquid_legions_test.py
```diff
@@ -0,0 +1,141 @@
...
+)
+from wfa_cardinality_estimation_evaluation_framework.estimators.same_key_aggregator import (
+    ExponentialSameKeyAggregator,
+    StandardizedHistogramEstimator,
+)
+from wfa_cardinality_estimation_evaluation_framework.estimators.estimator_noisers import (
+    LaplaceEstimateNoiser,
+)
+
+reach_epsilon = 0.1  # dp epsilon for reach
```
matthewclegg, line 33: Consider moving this code into a setUpClass() method.

https://docs.python.org/3/library/unittest.html#setupclass-and-teardownclass

src/accuracy/tests/analytical_variance_for_sampling_liquid_legions_test.py
```diff
@@ -0,0 +1,141 @@
+# Copyright 2021 The Private Cardinality Estimation Framework Authors
```
matthewclegg, line 1: Can you reduce the amount of time needed for these tests to execute?

src/accuracy/tests/analytical_variance_for_sampling_liquid_legions_test.py
```diff
@@ -0,0 +1,141 @@
...
+        """Test whether the stds obtained by analytical formulas roughly aligh with simulation.
+
+        Args:
+            n:  True 1+ reach.
+            relative_hist:  True relative frequency histogram.
+        """
+
+        def test_theory_discrepency(empirical, theoretical, type):
+            self.assertTrue(
+                (abs(empirical - theoretical) / theoretical < 0.5)
```
matthewclegg, line 117: Just checking ... a 50% relative error is considered acceptable?  Perhaps it would be good to include some tests to ensure that the following functions compute the expected results on some specific, known values: var_n_hat, std_n_hat, relative_std_n_hat, get_ea, var_rk_hat, std_rk_hat.

src/accuracy/tests/analytical_variance_for_sampling_liquid_legions_test.py
```diff
@@ -0,0 +1,141 @@
...
+    )
+    def test_alignment_with_empirical(self, n, relative_hist):
+        """Test whether the stds obtained by analytical formulas roughly aligh with simulation.
+
+        Args:
+            n:  True 1+ reach.
+            relative_hist:  True relative frequency histogram.
+        """
+
+        def test_theory_discrepency(empirical, theoretical, type):
```
matthewclegg, line 115: discrepancy

PR #101, opened 2022-01-14
Compute kreach and knreach using broadcast

riemanli:
The code in `knreach` was not thread-safe and causing concurrency issue when executed with multi-threading. The root cause is that `scipy.stats.binom` is an object, so within the same interpreter, all the threads are using the same `scipy.stats.binom` object. The solution proposed in this PR is to create a new Scipy binom object every time we need to use binomial distribution.

In addition to the threading issue, we utilize broadcasting in Numpy to speed up the computation in `knreach`. The function is twice faster in the test.

Additional to the two main changes, this PR also includes several bug-fixes and modifications:
1. Trailing spaces in README
2. Use na_rep="NaN" in “to_csv” for missing data representation.
3. Using traceback message as the model exception message
4. Avoid empty list as an argument in max()
5. Use "None" when there is no model exception

### DIFF
--- a/src/driver/experimental_trial.py
+++ b/src/driver/experimental_trial.py
@@ -215,7 +215,7 @@ class ExperimentalTrial:
                 logging.vlog(1, f"Trial   {self._trial_descriptor}")
             logging.vlog(1, f"Modeling failure: {inst}")
             logging.vlog(2, traceback.format_exc())
-            metrics = aggregate_on_exception(inst)
+            metrics = aggregate_on_exception(traceback.format_exc())
             if self._analysis_type == SINGLE_PUB_ANALYSIS:
                 single_publisher_dataframe = (
                     self._single_publisher_fractions_dataframe_on_exception(
@@ -239,7 +239,9 @@ class ExperimentalTrial:
         filesystem.mkdir(
             filesystem.parent(trial_results_path), parents=True, exist_ok=True
         )
-        filesystem.write_text(trial_results_path, result.to_csv(na_rep="NaN", index=False))
+        filesystem.write_text(
+            trial_results_path, result.to_csv(na_rep="NaN", index=False)
+        )
         filesystem.unlink(pending_path, missing_ok=True)
 
         return result
--- a/src/driver/experimental_trial.py
+++ b/src/driver/experimental_trial.py
@@ -239,7 +239,7 @@ class ExperimentalTrial:
         filesystem.mkdir(
             filesystem.parent(trial_results_path), parents=True, exist_ok=True
         )
-        filesystem.write_text(trial_results_path, result.to_csv(index=False))
+        filesystem.write_text(trial_results_path, result.to_csv(index=False, na_rep="NaN"))
         filesystem.unlink(pending_path, missing_ok=True)
 
         return result
--- a/src/driver/experimental_trial.py
+++ b/src/driver/experimental_trial.py
@@ -218,7 +218,9 @@ class ExperimentalTrial:
             metrics = aggregate_on_exception(inst)
             if self._analysis_type == SINGLE_PUB_ANALYSIS:
                 single_publisher_dataframe = (
-                    self._single_publisher_fractions_dataframe_on_exception(max_frequency)
+                    self._single_publisher_fractions_dataframe_on_exception(
+                        max_frequency
+                    )
                 )
 
         independent_vars = self._make_independent_vars_dataframe()
@@ -338,15 +340,21 @@ class ExperimentalTrial:
         results["max_nonzero_frequency_from_halo"] = [
             max(
                 [(i + 1) for i, f in enumerate(training_point._kplus_reaches) if f != 0]
+                + [0]
             )
         ]
         data_point = halo.true_reach_by_spend(halo.campaign_spends, max_frequency)
         results["max_nonzero_frequency_from_data"] = [
-            max([(i + 1) for i, f in enumerate(data_point._kplus_reaches) if f != 0])
+            max(
+                [(i + 1) for i, f in enumerate(data_point._kplus_reaches) if f != 0]
+                + [0]
+            )
         ]
         return pd.DataFrame(results)
 
-    def _single_publisher_fractions_dataframe_on_exception(self, max_frequency) -> pd.DataFrame:
+    def _single_publisher_fractions_dataframe_on_exception(
+        self, max_frequency
+    ) -> pd.DataFrame:
         results = {}
         for r in SINGLE_PUBLISHER_FRACTIONS:
             column_name = f"relative_error_at_{int(r*100):03d}"
--- a/src/driver/experimental_trial.py
+++ b/src/driver/experimental_trial.py
@@ -339,19 +339,26 @@ class ExperimentalTrial:
         # Also, record the maximum frequency in the actual data and the
         # data produced by Halo.
         training_point = reach_surface._data[0]
-        results["max_nonzero_frequency_from_halo"] = [
-            max(
-                [(i + 1) for i, f in enumerate(training_point._kplus_reaches) if f != 0]
-                + [0]
-            )
-        ]
+        results["max_nonzero_frequency_from_halo"] = (
+            [
+                max(
+                    [
+                        (i + 1)
+                        for i, f in enumerate(training_point._kplus_reaches)
+                        if f != 0
+                    ]
+                )
+            ]
+            if training_point._kplus_reaches
+            else [0]
+        )
         data_point = halo.true_reach_by_spend(halo.campaign_spends, max_frequency)
-        results["max_nonzero_frequency_from_data"] = [
-            max(
-                [(i + 1) for i, f in enumerate(data_point._kplus_reaches) if f != 0]
-                + [0]
-            )
-        ]
+        results["max_nonzero_frequency_from_data"] = (
+            [max([(i + 1) for i, f in enumerate(data_point._kplus_reaches) if f != 0])]
+            if data_point._kplus_reaches
+            else [0]
+        )
+
         return pd.DataFrame(results)
 
     def _single_publisher_fractions_dataframe_on_exception(
--- a/src/driver/experimental_trial.py
+++ b/src/driver/experimental_trial.py
@@ -239,7 +239,7 @@ class ExperimentalTrial:
         filesystem.mkdir(
             filesystem.parent(trial_results_path), parents=True, exist_ok=True
         )
-        filesystem.write_text(trial_results_path, result.to_csv(index=False, na_rep="NaN"))
+        filesystem.write_text(trial_results_path, result.to_csv(na_rep="NaN", index=False))
         filesystem.unlink(pending_path, missing_ok=True)
 
         return result
--- a/src/driver/test_point_aggregator.py
+++ b/src/driver/test_point_aggregator.py
@@ -140,7 +140,7 @@ def aggregate(
     Returns:
       A single row DataFrame representing the values of the statistics listed in keys.
     """
-    stats = {"model_succeeded": [1], "model_exception": [""]}
+    stats = {"model_succeeded": [1], "model_exception": ["None"]}
     for key in AGGREGATORS:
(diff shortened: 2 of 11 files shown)


src/driver/test_point_aggregator.py
```diff
@@ -140,21 +140,22 @@ def aggregate(
     Returns:
       A single row DataFrame representing the values of the statistics listed in keys.
     """
-    stats = {"model_succeeded": [1], "model_exception": [""]}
+    stats = {"model_succeeded": [1], "model_exception": ["None"]}
```
matthewclegg, line 143: Why this change?

src/driver/test_point_aggregator.py
```diff
@@ -140,21 +140,22 @@ def aggregate(
...
 
     Args:
-      inst:  The exception instance that was generated in the modeling attempt.
+      debug_msg:  The message for debugging the cause of the exception that was
+        generated in the modeling attempt.
     Returns:
       A single row DataFrame of NAs with columns being the statistics listed in keys.
     """
-    stats = {"model_succeeded": [0], "model_exception": [str(inst)]}
+    stats = {"model_succeeded": [0], "model_exception": [debug_msg]}
```
matthewclegg, line 158: Is there a limit on the length of a line in a CSV file or on the length of a field?

src/models/gamma_poisson_model.py
```diff
@@ -177,19 +177,23 @@ def _knreach(self, k, n, I, Imax, alpha, beta):
...
           beta: Parameter of Gamma-Poisson distribution.
         Returns:
-          Probability that a randomly chosen user will have an inventory
-          of n impressions, of which k are shown.
+          (C, M) ndarray. Probability that a randomly chosen user will have an
+          inventory of n impressions, of which k are shown.
         """
-        kprob = scipy.stats.binom.logpmf(k, n, I / Imax)
-        return np.exp(kprob + self._logpmf(n, alpha, beta))
+        binom_dist = scipy.stats._discrete_distns.binom_gen(name="binom_dist")
```
matthewclegg, line 190: Did you test the idea of creating a binom_dist object in the constructor?

matthewclegg (commented):
Thanks, Rieman!

src/driver/test_point_aggregator.py
riemanli:
```python
result.to_csv(na_rep="NaN")
```
This will make not only `np.NaN` but also empty cells to `"NaN"`. To avoid the wrong conversion for the empty cells in `model exception`, I put "None" into the cells instead.
I think this is reasonable because an empty cell in a `dataframe` usually means missing data. Every cell should have a clear value.

src/driver/test_point_aggregator.py
riemanli: Not for data storage. The only limitation is that if you are reading the .csv file with Excel, Excel can only view ~32k chars per cell.

src/models/gamma_poisson_model.py
riemanli: Yes. I just got the result, and it looks good. I'll share the result with you and change the code to use `self.binom_dist`.

src/driver/test_point_aggregator.py
matthewclegg: sgtm

src/driver/test_point_aggregator.py
matthewclegg: ok,thanks for checking

matthewclegg (approved):

riemanli:
Hi @amiragha , just a kind reminder of this review. Please let me know if you have any questions. Thanks!

src/models/gamma_poisson_model.py
```diff
@@ -201,30 +206,22 @@ def _kreach(self, k, I, Imax, alpha, beta):
...
-                )
-                for kv in k
-            ]
-        )
+        k = np.asarray(k)
+        mat = self._knreach(
+            k, np.arange(MAXIMUM_COMPUTATIONAL_FREQUENCY + 1), I, Imax, alpha, beta
+        )  # shape: C x MAXIMUM_COMPUTATIONAL_FREQUENCY
+        upper_triangular = np.triu(mat)
+        return np.sum(upper_triangular, axis=-1)
```
amiragha, line 224: Can we not get the same behavior using the `numpy.cumsum`?

The choice of the upper triangular matrix and summing over is generally confusing here IMO, I suggest adding a comment describing what is going on.

amiragha (approved):
@riemanli  Sorry I forgot this PR for so long and thanks for the reminder.

This looks good to me, I approve and add a few suggestions.

src/driver/experimental_trial.py
```diff
@@ -338,15 +342,21 @@ def _compute_single_publisher_fractions_dataframe(
         results["max_nonzero_frequency_from_halo"] = [
             max(
                 [(i + 1) for i, f in enumerate(training_point._kplus_reaches) if f != 0]
+                + [0]
```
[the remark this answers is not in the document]
riemanli: Thanks for the suggestion! I will modify the code.

src/models/gamma_poisson_model.py
riemanli: The changes utilize broadcasting and matrix multiplication with upper triangle matrix to speed up the computation. I am not sure if using `numpy.cumsum` can achieve the same result and performance.

src/driver/experimental_trial.py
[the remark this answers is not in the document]
riemanli: Modified. @amiragha Please take a look. Thank you!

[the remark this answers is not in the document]
amiragha: Thanks

src/models/gamma_poisson_model.py
amiragha: @riemanli Here, I only see a sum of with the upper triangular matrix over the last axis. Where is the matrix multiplication with the upper triangular that you mentioned, can you point it out to me?

riemanli: Ah I see. Sorry I misunderstood your points. Please neglect my statement above. Let me answer it again.

> Can we not get the same behavior using the `numpy.cumsum`?

I am assuming you're asking about line 224 only. Given that, I don't think the behavior of `np.sum(axis=-1)` is the same as `np.cumsum(axis=-1)`. Please let me know if I misunderstand something again.

> The choice of the upper triangular matrix and summing over is generally confusing here IMO, I suggest adding a comment describing what is going on.

Sure. Let me explain it here as well. `mat` is a 2-dimension array, so `np.triu` will generate `upper_triangular` as a 2d array from 'mat' with all elements below diagonal in 'mat' being zeros. Then the result is the sum of `upper_triangular` over the last axis and is a 1d array. Ex:
```
mat = 
[
  [1, 2],
  [3, 4],
  [5, 6],
]
upper_triangular = 
[
  [1, 2],
  [0, 4],
  [0, 0],
]
result = 
  [3, 4, 0]
```

amiragha: @riemanli

Thank you for then detailed explanation! Could you please also explain why the new code is equivalent to the previous version? This is not clear to me. I might be missing something here...

amiragha (approved):
I approve so that go ahead and merge.

However, I don't think I quite understand how using the upper triangular matrix and sum is equivalent to the previous code (which only uses sum) and would appreciate a comment (maybe a comment in the actual code too) in order to clarify this. Thanks a lot!

src/models/gamma_poisson_model.py
riemanli: For sure. I would love to explain the rationale behind it.

If we look at the original code of `_kreach`
```python
return np.array(
    [
        np.sum(
            self._knreach(
                kv,
                np.arange(kv, MAXIMUM_COMPUTATIONAL_FREQUENCY + 1),
                I,
                Imax,
                alpha,
                beta,
            )
        )
        for kv in k
    ]
)
```
and the original code of `_knreach` (note that I switch `k` to `kv` for consistence here):
```python
kprob = scipy.stats.binom.logpmf(kv, n, I / Imax)
return np.exp(kprob + self._logpmf(n, alpha, beta))
```

As a result, the result can be see as the pseudo code below
```python
M = MAXIMUM_COMPUTATIONAL_FREQUENCY    # M > C
k = [1, ..., C]
result = [
    np.sum(
        np.exp(
            scipy.stats.binom.logpmf(kv = 1, np.array([kv = 1, ..., M]), I / Imax) + 
            self._logpmf(np.array([kv = 1, ..., M]), alpha, beta)
        )
    ),
    .
    .
    np.sum(
        np.exp(
            scipy.stats.binom.logpmf(kv = C, np.array([kv = C, ..., M]), I / Imax) + 
            self._logpmf(np.array([kv = C, ..., M]), alpha, beta)
        )
    )
] # size = (C, )
```

So here is what the new change did. Since every element has `np.sum` operation, we can defer it later. Instead of using for-loop, I treat every element as a sum of array with size (1, M). As a result, we can first compute a 2D array `mat`
``` python
mat = [
    [_knreach(kv=1, n=0, other_args), ..., _knreach(kv=1, n=M, other_args)],
    [_knreach(kv=2, n=0, other_args), ..., _knreach(kv=1, n=M, other_args)],
    .
    .
    [_knreach(kv=C, n=0, other_args), ..., _knreach(kv=1, n=M, other_args)]
]
```
Since n is always greater than or equal to kv, we don't need the element with `kv < n`. As a result, we only need the upper triangle part of `mat`. And the last step is taking sum on each row of the upper triangle matrix.

The beauty of this change is that it can utilize the powerful broadcast in `Numpy`, which saves a lot computation time.

Hope that helps!

amiragha: @riemanli 
Awesome, thanks a lot for walking me through the details. This makes sense to me. 

One last questions:
Why is this guaranteed that the input vector k has this structure : `k = [0, ..., C - 1]`?

riemanli: Thanks for putting effort on it! That's a great question. The usage of `_knreach` is to compute the reach frequency distribution from the Gamma Poisson model. The new implementation assumes the values in `k` are consecutive frequency values, i.e. `k = np.arange(min_freq, max_freq + 1)`, which is a fair assumption given that we need the information from every frequency to fit the model well.

If there is any case that we want to have non-consecutive frequency values, e.g. `k = [0, 2, 4, ..., max_freq]`,  we can still generate the distribution from 0 to max_freq and extract the values at [0, 2, 4, ..., max_freq]. The computation won't be wasted too much in this approach as `max_freq` is usually small and broadcasting is fast.

amiragha: @riemanli 

I think your explanation of why you are using the upper triangular matrix is only valid when `k = [0, ..., C - 1]`. Therefore since you mentioned the possibility of `k = np.arange(min_freq, max_freq + 1)` and since it is technically possible for other inputs, I advise against your approach in this PR.

However, if you still prefer to keep it as is, please add to the documentation that `k` has to have a required form and possibly even test for the input to follow that form.

Thanks a lot for following up during this relatively long conversation. I appreciate it.

riemanli: Thank you so much for the thorough review @amiragha! 

@matthewclegg please let me know if you also think we should not have the assumption that `k = np.arange(min_freq, max_freq + 1)` for `_kreach`.

matthewclegg: Since it's a private function and since Rieman's implementation is consistent with how the function is being used within the class, I think it's ok.

amiragha (approved):

src/models/gamma_poisson_model.py
riemanli: Thanks for your reply Matthew.

@amiragha I added some comments and unit tests for the assumption about `k` as per your suggestion. Please let me know if there is any advice. Big thanks.

amiragha: @riemanli Great, thanks a lot for adding the comments as well as walking me through the details.

amiragha (approved):

PR #102, opened 2022-02-14
Dirac mixture single and multi pub models using grid-type search (Review will be done in PRs 103-106 separately)

jiayu-google:
I decided to put the single and multi pub models in the same PR because they are sharing many common utils. 

@matthewclegg  I will close the previous PR 99.  Your suggestions in PR 99 have been received and addressed in this PR.  Thank you!

### DIFF
--- a/src/models/dirac_mixture_multi_publisher_model.py
+++ b/src/models/dirac_mixture_multi_publisher_model.py
@@ -1,0 +1,579 @@
+# Copyright 2021 The Private Cardinality Estimation Framework Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,579 @@
+# Copyright 2021 The Private Cardinality Estimation Framework Authors
```
matthewclegg, line 1: Hi Jiayu,

This looks very interesting.  Thank you for writing it.  This is going to be a major effort to review this code, so I would be very grateful if you could break it into two or more PR's.  Also, is there an updated doc that fully describes the current version of the models?  This would be very helpful in reviewing the code.  Thanks!

Matt

matthewclegg (changes requested):

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Leaving a note: following Matt's suggestion, this PR is split into 4 PRs for review, tentatively PR 103, 105, 106, 107.  You may still read this PR for the complete codes, but please go to split PRs for conversations and updates.

PR #107, opened 2022-03-16
Change git: to https: for the new protocol change in github.

riemanli:


jiayu-google (approved):
Bravo, thanks a ton Rieman!

PR #108, opened 2022-04-01
CopulaDataSet to generate synthetic data.

jiayu-google:
Next PR: Modify `data_set_parameters` and `m3_data_design` to incorporate the copula synthetic data generators.


src/data_generators/copula_data_set.py
```diff
@@ -0,0 +1,205 @@
...
+from statsmodels.distributions.copula.elliptical import GaussianCopula
+
+
+from wfa_planning_evaluation_framework.data_generators.publisher_data import (
+    PublisherData,
+)
+from wfa_planning_evaluation_framework.data_generators.data_set import (
+    DataSet,
+)
+from wfa_planning_evaluation_framework.data_generators.data_set import (
```
chenweiw, line 33: The codes in lines 33~35 repeat those in lines 30~32.

chenweiw: The codes in lines 33-35 repeat those in lines 30-32.

src/data_generators/copula_data_set.py
```diff
@@ -0,0 +1,205 @@
...
+        Returns:
+            A dict d where for any frequency vector (f_1, ..., f_p),
+            d[(f_1, ..., f_p)] is the count of this frequency vector in
+            the DataSet.
+        """
+        return dict(Counter(tuple(obs) for obs in self.sample))
+
+    @staticmethod
+    def zero_included_pmf(
+        publisher_data: PublisherData, universe_size: int
```
chenweiw: The integer type is "int32" in line 131 and "int" in line 158. While there is no difference in running, shall we use it in a consistent manner?

src/data_generators/copula_data_set.py
```diff
@@ -0,0 +1,205 @@
...
+    def ppf(self, p: float) -> int:
+        """Calculates ppf of the count distribution defined by the histogram.
+
+        ppf is the inverse of cdf as will be explained in "Returns".
+
+        Args:
+            p:  A probability.
+
+        Returns:
+            A count x for which cdf[x] <= p and cdf[x + 1] > p, where cdf is
```
amiragha: This may not be satisfied if a frequency bucket has zero.

src/data_generators/tests/copula_data_set_test.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def test_to_impressions(self):
+        frequnecy_vectors = [(1, 1), (1, 1), (2, 0), (2, 0), (3, 3)]
+        frequnecy_vectors = [np.array(vec) for vec in frequnecy_vectors]
+        res = CopulaDataSet.to_impressions(frequnecy_vectors)
+        expected = [[0, 1, 2, 3, 2, 3, 4, 4, 4], [0, 1, 4, 4, 4]]
+        np.testing.assert_equal(res[0], expected[0])
+        np.testing.assert_equal(res[1], expected[1])
+
+    def test_approximate_agreeement_with_marginals(self):
```
riemanli, line 61: typo: agreeement

src/data_generators/tests/copula_data_set_test.py
```diff
@@ -0,0 +1,145 @@
...
+            universe_size=300,
+            random_generator=np.random.default_rng(0),
+        )
+
+        def frequency_dictionary(pdf: PublisherData) -> Dict:
+            freq_by_vid = pdf.user_counts_by_impressions(pdf.max_impressions)
+            return dict(Counter(freq_by_vid.values()))
+
+        res1, res2 = [frequency_dictionary(pdf) for pdf in dataset._data]
+        try:
```
riemanli, line 78: Why do we use `try-except` here?

src/data_generators/tests/synthetic_data_design_generator_test.py
```diff
@@ -47,6 +54,22 @@
...
 
+TEST_LEVELS_COPULA = {
+    "largest_publisher_size": [8, 16],
+    "overlap_generator_params": [
+        GeneratorParameters(
+            "Copula",
+            CopulaDataSet,
+            {
+                "largest_pub_to_universe_ratio": 0.5,
+                "copula_generator": GaussianCopula(corr=0),
```
riemanli, line 65: Would be better if you could use values other than the default value in `CopulaDataSet`.

src/data_generators/tests/copula_data_set_test.py
jiayu-google: Thanks!

src/data_generators/tests/copula_data_set_test.py
jiayu-google: In the very very rare case, one of the keys (freq = 1, 2, or 3) may not exist.  It's like we might have `1 not in res` instead of `res[1] = 0`.

src/data_generators/tests/synthetic_data_design_generator_test.py
jiayu-google: Good point.  Changed.

jiayu-google (commented):
Thank you @riemanli for the very careful review!

src/data_generators/tests/copula_data_set_test.py
riemanli: If that's something wrong, we can use `assertTrue` to capture it instead
```python
self.assertTrue(1 in res1)
self.assertTrue(2 in res1)
self.assertTrue(3 in res2)
```

riemanli: If that's something wrong, we can use assertTrue to capture it instead
```python
self.assertTrue(1 in res1)
self.assertTrue(2 in res1)
self.assertTrue(3 in res2)
```

src/driver/tests/experimental_trial_test.py
```diff
@@ -432,7 +432,7 @@ def test_evaluate_single_publisher_model(self):
             # been checked by the preceding unit tests.  However, we make a few strategic
             # probes.
             self.assertEqual(result.shape[0], 1)
-            self.assertAlmostEqual(result["relative_error_at_100"][0], 0.0, delta=0.01)
+            self.assertAlmostEqual(result["relative_error_at_100"][0], 0.0, delta=0.02)
```
riemanli, line 435: I don't know about the computation behind "relative_error_at_100". Could you confirm that the increase of `delta` is due to computational error but not some logical error?

jiayu-google: Yes I indeed should check it. 

1. In the main branch with original codes, I varied the `seed` in `trial.evaluate` above.  For `seed in range(200`, 6 seeds lead to error > 0.01 (and <0.02).  So, it is a computational error and delta=0.02 is reasonable.
2. This test does not use Copula generator at all, so is not affected by this PR except for the randomness.

PLMK what you think.

jiayu-google (commented):
Thanks Rieman! Please take another look.

src/driver/tests/experimental_trial_test.py
riemanli: Thanks for checking. That sounds good to me.

riemanli (approved):
Thanks Jiayu!

jiayu-google:
Thanks Rieman!

@amiragha @ecancellieri It'll be great to have your comments on this PR, especially since you've contributed a lot to this! 
Our next PR will be the parameter settings based on this PR.

src/data_generators/copula_data_set.py
```diff
@@ -0,0 +1,207 @@
...
+)
+from wfa_planning_evaluation_framework.data_generators.pricing_generator import (
+    PricingGenerator,
+)
+from wfa_planning_evaluation_framework.data_generators.fixed_price_generator import (
+    FixedPriceGenerator,
+)
+
+
+class AnyFrequencyDistribution:
```
matthewclegg: Why not use scipy.stats.rv_histogram?

src/data_generators/copula_data_set.py
```diff
@@ -0,0 +1,207 @@
...
+
+        Args:
+            unlabeled_publisher_data_list:  A list of PublisherData. Each PublisherData
+                indicates the frequency distribution at a publisher --- the
+                ids labels are meaningless while their distribution matters.
+            copula_generator:  An instance of a subclass of
+                statsmodels.distributions.copula.copulas.Copula.  See here for the
+                possible choices:
+                https://github.com/statsmodels/statsmodels/tree/32dc52699f994acbf9dbdb9bd10d7eff04d860f5/statsmodels/distributions
+            universe_size:  A cross-publisher universe size to construct the copula.
```
matthewclegg, line 64: Explain what happens if no universe size is given.

src/data_generators/copula_data_set.py
jiayu-google: Good to learn that `int` also works in astype, Chenwei.   After evaluating the trade-off, I propose to keep the current codes:
1. astype(int) defaults int64, while int32 is enough here.  So, keeping astype(int32) below.
2. As for typing, I never saw a case of specifying int32 etc. in the typing.  Typing does not have to be that strict.
PLMK what you think!

src/data_generators/copula_data_set.py
```diff
@@ -0,0 +1,207 @@
...
+            marginals=self.marginal_distributions,
+        )
+        self.sample = self.distribution.rvs(
+            nobs=universe_size,
+            random_state=random_generator.integers(0, 1e9),
+        ).astype("int32")
+        super().__init__(
+            publisher_data_list=[
+                PublisherData(
+                    impression_log_data=pricing_generator(pub_imps),
```
matthewclegg, line 99: Should you first shuffle pub_imps?

matthewclegg (changes requested):

src/data_generators/copula_data_set.py
jiayu-google: Thanks for finding this!
`scipy.stats.rv_histogram` is powerful, but there's one concern.  As can be seen from its docstring, it is a subclass of `rv_continuous`.  It's constructing a continuous variable from a histogram, and not straightforwardly constructing a discrete variable.

A specific example of its difference from `AnyFrequencyDistribution`: 
https://screenshot.googleplex.com/7dqsiJRi8ucFyWn

With this difference, please feel free to let me know if you think rv_histogram makes more sense than AnyFrequencyDistribution!

src/data_generators/copula_data_set.py
jiayu-google: Thanks, done.

src/data_generators/copula_data_set.py
jiayu-google: You are right.  Thanks for catching this important bug!
Added shuffle in lines 129-131

jiayu-google (commented):
Thanks Matt.  pt another look.

src/data_generators/copula_data_set.py
matthewclegg: What about scipy.stats.rv_discrete?

matthewclegg (changes requested):

src/data_generators/copula_data_set.py
jiayu-google: Yes, it works!  Updated.

matthewclegg (approved):

src/data_generators/copula_data_set.py
```diff
@@ -0,0 +1,176 @@
...
+    @property
+    def frequency_vectors_sampled_distribution(self) -> Dict:
+        """A dictionary of the distribution of the sampled frequency vectors.
+
+        Returns:
+            A dict d where for any frequency vector (f_1, ..., f_p),
+            d[(f_1, ..., f_p)] is the count of this frequency vector in
+            the DataSet.
+        """
+        return dict(Counter(tuple(obs) for obs in self.sample))
```
amiragha, line 119: Just curious, what does "obs" stand for here? Observation?

src/data_generators/copula_data_set.py
```diff
@@ -0,0 +1,176 @@
...
+
+    @staticmethod
+    def zero_included_pmf(
```
amiragha: Suggestion: Shouldn't this be a method of `PublisherData` class instead?

creating a histogram of frequencies starting from zero can be useful other place other then `CopulaDataSet` class.

src/data_generators/copula_data_set.py
```diff
@@ -0,0 +1,176 @@
...
+        self.distribution = CopulaDistribution(
+            copula=self.copula_generator,
+            marginals=self.marginal_distributions,
+        )
+        self.sample = self.distribution.rvs(
+            nobs=universe_size,
+            random_state=random_generator.integers(0, 1e9),
+        ).astype("int32")
+        imps = self.to_impressions(self.sample)
+        for pub_imps in imps:
```
amiragha: suggestion: The shuffling that Matthew suggested can also be done inside or the `to_impression` method

src/data_generators/overlap_data_set.py
```diff
@@ -30,11 +30,15 @@
...
 
 class OverlapDataSet(DataSet):
-    """Construct a multi-pub DataSet from a list of single-pub PublisherData.
+    """A DataSet constructed from a list of PublisherData by overlapping their reached sets.
 
     Once we have a list of PublisherData for each publisher, this class
     generates the cross-publisher reach overlap, relabels the reached ids to
     reflect the overlap, and finally includes these labeled ids to a DataSet.
+
+    In other words, the universes are overlapped while for each user, their
```
amiragha: The wording here is not clear to me. I didn't understand what it means! Do you mean that condition on having a fixed overlap for universes as input, then the frequency distribution is generated?

amiragha (approved):
@jiayu-google  

Sorry for very late review. Great work. I just added some suggestions, feel free to ignore.

src/data_generators/copula_data_set.py
jiayu-google: Makes sense.  Now moved the shuffling to `to_impressions`.

src/data_generators/copula_data_set.py
jiayu-google: Yep, observation

src/data_generators/copula_data_set.py
jiayu-google: Great suggestion.  This is indeed needed elsewhere (for example, in the future DiracMixtureDataSet).

src/data_generators/overlap_data_set.py
jiayu-google: Yes.  Thanks Amir, your wording is clearer.  I've expanded your wording a little bit.

jiayu-google (commented):
Hi @amiragha ,
Thanks for your constructive suggestions!  

Sorry, the last weeks were crazy for me, I didn't get a chance to finalize this revision.  Please take another look when you have time.

PR #110, opened 2022-04-05
Generator of the subset test points.

jiayu-google:


src/driver/m3_subset_test_point_generator.py
```diff
@@ -0,0 +1,110 @@
...
+    TestPointGenerator,
+)
+
+
+class M3SubsetTestPointGenerator(TestPointGenerator):
+    """Generates test points for the subset use case with the m3 strategy.
+
+    In M3, the training points are the frequency histograms of
+    - each of the single publishers,
+    - each of the all-but-one subsets of publishers,
```
chenweiw, line 31: If I have never worked on this, I would be stuck at understanding the meaning of "the frequency histogram of an all-but-one subset".

src/driver/m3_subset_test_point_generator.py
```diff
@@ -0,0 +1,110 @@
...
+    ):
+        """Constructs an M3SubsetTestPointGenerator.
+
+        Args:
+            dataset:  The DataSet for which test points are to be generated.
+            campaign_spend_fractions:  The campaign_spend / inventory_spend
+                at each publisher, where each inventory_spend is given by
+                dataset._max_spends.
+            max_num_points:  If (2^p - 1) - (2p + 1) > max_num_points, then
+                randomly sample max_num_points subsets.
```
chenweiw: -> randomly sample max_num_points subsets from the (2^p-1) - (2p+1) subsets.

src/driver/m3_subset_test_point_generator.py
```diff
@@ -0,0 +1,110 @@
...
+        self.max_num_points = max_num_points
+        self._campaign_spends = self._max_spends * campaign_spend_fractions
+
+    @staticmethod
+    def subset_layer(p: int, q: int) -> List[np.ndarray]:
+        """Returns a list of subset indicators for a layer of subsets.
+
+        A layer of subsets means the collection of subsets with the same
+        cardinality.
+
```
chenweiw, line 79: I think maybe "subset indicator" also needs an explanation here, as it is not a widely used phrase.

src/driver/m3_subset_test_point_generator.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        p = self._npublishers
+        if p < 4:
+            return
+        if (2 ** p - 1) - (2 * p + 1) > self.max_num_points:
+            n = 0
+            while n < self.max_num_points:
+                subset_indicator = np.random.choice([0, 1], p)
+                num_involved_subsets = np.count_nonzero(subset_indicator)
+                if num_involved_subsets > 1 and num_involved_subsets < p - 1:
```
chenweiw: It is equivalent to "1 < num _involved_subsets < p-1".

src/driver/m3_subset_test_point_generator.py
```diff
@@ -0,0 +1,110 @@
...
+            the true reach surface is to be compared to the modeled reach
+            surface.
+        """
+        p = self._npublishers
+        if p < 4:
+            return
+        if (2 ** p - 1) - (2 * p + 1) > self.max_num_points:
+            n = 0
+            while n < self.max_num_points:
+                subset_indicator = np.random.choice([0, 1], p)
```
chenweiw: Here, random.choice with the default replace = True is used, and thus the sampled self.max_num_points points could have repetition among themselves. Shall we need to guarantee that the self.max_num_points points have no repetition?

src/driver/m3_subset_test_point_generator.py
jiayu-google: Thanks for letting me know.  Which part do you think is difficult to understand?  The meaning of "all-but-one subset", or how to calculate the frequency histogram of it?

src/driver/m3_subset_test_point_generator.py
jiayu-google: Done.

src/driver/m3_subset_test_point_generator.py
jiayu-google: Good point.  I've improved the description of this method, ptal.

src/driver/m3_subset_test_point_generator.py
jiayu-google: Each subset_indicator is a length <p> binary vector.  So, I'm not really doing "choice" here, but just generating a random binary vector.

src/driver/m3_subset_test_point_generator.py
jiayu-google: Yep, simplified!

jiayu-google (commented):
Please take another look, Chenwei!

src/driver/m3_subset_test_point_generator.py
```diff
@@ -0,0 +1,118 @@
...
+                num_involved_subsets = np.count_nonzero(subset_indicator)
+                if 1 < num_involved_subsets < p - 1:
+                    yield list(self._campaign_spends * subset_indicator)
```
matthewclegg: How do you ensure that each of the test points generated by this method is distinct?

matthewclegg (changes requested):

src/driver/m3_subset_test_point_generator.py
jiayu-google: It's not ensured.  Some thoughts
1. It may be fine to allow the same test points?
2. If we want to ensure they are distinct, I can change the code using, say, `set`.  This adds computational complexity though.
PLMK what you think!

matthewclegg: I think the points should be distinct.

jiayu-google: Sure.  Just added a naive way to make them distinct, ptal.

src/driver/m3_subset_test_point_generator.py
chenweiw: I mean its meaning. On the other hand, I think it is very minor after a second thought. :)

chenweiw (approved):

matthewclegg (approved):

PR #111, opened 2022-04-05
Generators of the ShareShift test points.

jiayu-google:


src/driver/shareshift_test_point_generator.py
```diff
@@ -0,0 +1,116 @@
...
+        campaign_spend_fractions: np.ndarray,
+        shift_fraction_choices: List[float] = [-1.0, -0.5, 0.5, 1.0],
+    ):
+        """Constructs a ShareShiftTestPointGenerator.
+
+        Suppose that the original spend vector is [s_1, ..., s_p].
+        This class generates p * m testing points v with:
+            v[i] = s_i + s_i * a_k and
+            v[j] = s_j - s_j / (sum(s) - s[i]) * (s_i * a_k) for j != i,
+        iterating over 1 <= i <= p and k = 1, ..., m.  Here, a_1, ..., a_m are a list
```
chenweiw: [a_1, ..., a_m] is a list

src/driver/shareshift_test_point_generator.py
```diff
@@ -0,0 +1,116 @@
...
+            * [1 + (0.5 * 2) * 1 / (1 + 3), 2 - 0.5 * 2, 3 + (0.5 * 2) * 3 / (1 + 3)]
+            * [1 - (0.5 * 2) * 1 / (1 + 3), 2 + 0.5 * 2, 3 - (0.5 * 2) * 3 / (1 + 3)]
+            * [1 + (0.5 * 3) * 1 / (1 + 2), 2 + (0.5 * 3) * 2 / (1 + 2), 3 - 3 * 0.5]
+            * [1 - (0.5 * 3) * 1 / (1 + 2), 2 - (0.5 * 3) * 2 / (1 + 2), 3 + 3 * 0.5]
+
+        Args:
+            dataset:  The DataSet for which test points are to be generated.
+            campaign_spend_fractions:  The campaign_spend / inventory_spend
+                at each publisher, where each inventory_spend is given by
+                dataset._max_spends.
```
chenweiw: Similar to what Rieman recently suggested, probably add the dimension of each input argument?

src/driver/shareshift_test_point_generator.py
```diff
@@ -0,0 +1,116 @@
...
+            The spend vector of the testing point.  Explicitly, this vector
+            v has:
+                v[i] = s_i - s_i * a and
+                v[j] = s_j + s_j / (sum(s) - s[i]) * (s_i * a) for j != i,
+            where i = selected_pub and a = shift_fraction.
+        """
+        v = self._campaign_spends.copy()
+        shift_amount = v[selected_pub] * shift_fraction
+        if sum(v) - v[selected_pub] > 0:
+            shift_proportions = v / (sum(v) - v[selected_pub])
```
chenweiw, line 113: Does this equation for  um(v) - v[selected_pub] > 0 first appear? Based on the description in Returns, there seems no this definition?

src/driver/shareshift_test_point_generator.py
```diff
@@ -0,0 +1,116 @@
...
+        else:
+            shift_proportions = np.array(
+                [1 / (self._npublishers - 1)] * self._npublishers
+            )
+        v += shift_proportions * shift_amount
+        v[selected_pub] = self._campaign_spends[selected_pub] - shift_amount
```
chenweiw, line 122: After carefully reading, I can tell the codes are correct. The only issue is the codes might not be very easy to read.

src/driver/shareshift_test_point_generator.py
```diff
@@ -0,0 +1,116 @@
...
+        Returns:
+            An iterable of spend vectors representing locations where
+            the true reach surface is to be compared to the modeled reach
+            surface.
+        """
+        for selected_pub in range(self._npublishers):
+            for shift_fraction in self._shift_fraction_choices:
+                point = self.one_testing_point(selected_pub, shift_fraction)
+                if not any(point < 0):
+                    yield point
```
chenweiw, line 137: Based on the definition in the previous function, the entries in point can be positive or negative. But here, you only care about the cases where all the entries of point are non-negative. Am I right?

src/driver/shareshift_test_point_generator.py
```diff
@@ -0,0 +1,116 @@
...
+        dataset: DataSet,
+        campaign_spend_fractions: np.ndarray,
+        shift_fraction_choices: List[float] = [-1.0, -0.5, 0.5, 1.0],
+    ):
+        """Constructs a ShareShiftTestPointGenerator.
+
+        Suppose that the original spend vector is [s_1, ..., s_p].
+        This class generates p * m testing points v with:
+            v[i] = s_i + s_i * a_k and
+            v[j] = s_j - s_j / (sum(s) - s[i]) * (s_i * a_k) for j != i,
```
chenweiw: The notation here seems a bit confused. In particular, v is still a p-dimensional vector. Given an a_k, we obtain p entries, i.e., v and s have the same dimension. Then enumerating over a_1, a_2, ..., a_m, we obtain a total of p*m data points. It is better not to say p*m testing points v, because usually it means v has p*m entries.

jiayu-google: Thanks Chenwei.  Motivated by your comments, I've updated the description of this class.  Please take another look.

src/driver/shareshift_test_point_generator.py
jiayu-google: Good point, added.  And also revised the description of this arg.

src/driver/shareshift_test_point_generator.py
jiayu-google: Thanks, I've now called it our in the description.

src/driver/shareshift_test_point_generator.py
jiayu-google: Agreed.  Hope it's better now with the updated descriptions.

src/driver/shareshift_test_point_generator.py
jiayu-google: Exactly.  We discard test points with any negative spend.

jiayu-google (commented):
Thanks Chenwei!  I improved descriptions following your comments, ptal.

matthewclegg (approved):

chenweiw (approved):

PR #113, opened 2022-04-29
Dirac mixture multi pub model

jiayu-google:


jiayu-google:
Please let me know if you'd like to split this PR!

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+            UnivariateMixedPoissonOptimizer.validate_frequency_histogram(hist)
+        self.marginal_pmfs = self.normalize_rows(prior_marginal_frequency_histograms)
+        self.check_dimension_campatibility()
+
+        self.components = self.weighted_random_sampling(
+            ncomponents, self.marginal_pmfs, rng
+        )
+        self.fitted = False
+
+    def check_dimension_campatibility(self):
```
matthewclegg: change name to check_dimension_compatibility

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+            raise ValueError("Inconsistent number of directions")
+        # Minor note: for flexibity, we don't require self.marginal_pmfs to have
+        # the same max_freq as self.observed_pmf_matrix.
+
+    @classmethod
+    def normalize_rows(cls, matrix: np.ndarray) -> np.ndarray:
+        """Normalize a matrix so that each row sums up to 1."""
+        row_sums = matrix.sum(axis=1)
+        # Using numpy broadcasting
+        return matrix / row_sums[:, np.newaxis]
```
matthewclegg, line 124: Is it possible that a row sum can be zero?

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+
+    @classmethod
+    def weighted_random_sampling(
```
matthewclegg: The name "weighted_random_sampling" is perhaps a bit too generic.  Would it be better to use a name that specifies what is being sampled, or what the return value represents?

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+
+                As of Apr 2022, all the training points from halo_simulator will
+                have a binary observable_direction (like the above example).
+                We allow non-binary observable_direction in the future.
+            frequency_histograms_on_observable_directions:  An <n * (F + 1)> matrix
+                where n = #training points, F = maximum frequency.
+                Its i-th row is the observed frequency histogram at the i-th
+                row of observable_directions, i.e., the i-th training point.
+            prior_marginal_frequency_histograms:  A <p * (F + 1)> matrix where
+                n = #training points, F = maximum frequency.
```
chenweiw: delete "n = #training points" because there is no "n" in this argument.

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+
+                Its i-th row is a marginal frequency histogram at the i-th
+                publisher under the baseline impression (see the description of
+                the `observable_directions` arg).  In other words, its rows are
+                frequency_histograms at the single-pub observation_directions =
+                [1, 0, ..., 0], [0, 1, 0, ..., 0], ..., [0, ..., 0, 1],
+                respectively.  As of Apr 2022, halo_simulator generates a training
```
chenweiw: I personally feel that the sentence "In other words ... respectively", i.e., line 74-77, might makes the description a bit harder to interpret, because the vector [1, 0, .., 0] is with dimension p, corresponding to p publishers, which appeared in this class for the first time (while line 58 is for a special case). On the other hand, this is minor, and you can decide it. :)

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+        """Check the compatibility of dimensions among the matrix attributes."""
+        if self.marginal_pmfs.shape[0] != self.p:
+            raise ValueError("Inconsistent number of publishers")
+        if self.observable_directions.shape[0] != self.observed_pmf_matrix.shape[0]:
+            raise ValueError("Inconsistent number of directions")
+        # Minor note: for flexibity, we don't require self.marginal_pmfs to have
+        # the same max_freq as self.observed_pmf_matrix.
+
+    @classmethod
+    def normalize_rows(cls, matrix: np.ndarray) -> np.ndarray:
```
chenweiw, line 114: Consider adding the description and dimension of each input argument.

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+            rng:  Random generator.
+
+        Returns:
+            A <C * p> matrix where C = #components and p = #pubs.
+            Each row is a component, i.e., a vector of Poisson means at all pubs.
+        """
+        p, f = marginal_pmfs.shape
+        sample = np.array(
+            [rng.choice(a=range(f), p=pmf, size=ncomponents) for pmf in marginal_pmfs]
+        ) + rng.random(size=(p, ncomponents))
```
chenweiw: It might be better to add a comment for this line of codes, because "+ rng.random(size=(p, ncomponents))" seems not very intuitive, from my personal perspective.

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+                hypothetical_direction = [0.5, 1, 2] means that we want to predict
+                a hypothetical campaign with impression vector = [0.5 * 1e6,
+                1 * 2e6, 2 * 3e6].
+            customized_max_freq:  If specified, the predicted frequency
+                histogram will be capped by this maximum frequency. It can be
+                different from the maximum frequency of the training data.
+
+        Returns:
+            A length <F + 1> vector.
+            Its f-th coordinate is the pmf at frequency f of the hypothetical campiaign.
```
chenweiw: Consider adding "count from the zero index" right after "f-th coordinate".

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+        """
+        if len(hypothetical_direction) != self.p:
+            raise ValueError("Hypothetical direction does match number of publishers")
+        if not self.fitted:
+            self.fit()
```
chenweiw, line 299: There could be two options when self.fitted = False. Option 1: throw an error message, like the model has not been fitted. Option 2: it is what is implemented here. Although both options are correct, I feel usually Option 1 is more widely used. However, you can decide the way you like, afternoon there is no right or wrong.

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+        reach_curves: List[ReachCurve] = None,
+        single_publisher_reach_agreement: bool = True,
+        ncomponents: int = None,
+        rng: np.random.Generator = np.random.default_rng(0),
+    ):
+        """Constructs a Dirac mixture multi publisher model.
+
+        Args:
+            reach_points: A length <n> vector where n = #(training points).
+                List of ReachPoints on which the model is to be trained.
```
chenweiw, line 333: Since ReachPoint is not a very common type, what type is it?

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+            reach_curves: A length <p> list where p = #pubs.
```
chenweiw, line 334: Since ReachCurve is not a very common type, what type is it?

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+
+        Args:
+            reach_points:  Any list of ReachPoints.
+
+        Returns:
+            A dictionary of which the value of key = i is the sub-list of
+            reach points that involve only pub i.
+        """
+        single_pub_points = {}
+        p = len(reach_points[0].impressions)
```
chenweiw: Usually, the index [0] is used after getting rid of the corner case where reach_points is an empty list, considering the description "Any list of ReachPoints". We can simply add "if len(reach_points) == 0: return { }" or raise an error message at the top.

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+        for rp in reach_points:
+            involved_pubs = [i for i in range(p) if rp.impressions[i] > 0]
+            if len(involved_pubs) == 1:
+                pub = involved_pubs[0]
+                single_pub_points[pub] = (
+                    single_pub_points[pub] + [rp]
+                    if pub in single_pub_points.keys()
```
chenweiw: We can directly use "if pub in single_pub_points", i.e., without ".keys()".

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+                    else [rp]
+                )
```
chenweiw: lines 352-356 can be simplified to "single_pub_points[pub] = single_pub_points.get(pub, []) + [rp]"

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+        return single_pub_points
+
+    @classmethod
+    def obtain_marginal_frequency_histograms(
+        cls, reach_points: List[ReachPoint]
+    ) -> Tuple[np.ndarray]:
+        """One way to obtain prior_marginal_frequency_histograms as an input of MultivariateMixedPoissonOptimizer.
+
+        This and the following two functions are for translating the given
```
chenweiw: Better to specify which two?

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+            This method extracts the single pub frequeny histograms and put them in
+            a matrix, as the `prior_marginal_frequency_histograms` argument of
+            MultivariateMixedPoissonOptimizer.
+
+            The impressions at these single publisher points are also returned.
+            They will be used as baseline_impression_vector (see the description
+            of MultivariateMixedPoissonOptimizer).
+        """
+        single_pub_points = cls.select_single_publisher_points(reach_points)
+        p = len(reach_points[0].impressions)
```
chenweiw: Same as the above, shall we need to consider the corner case where reach_points = []?

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+        prior_marginal_frequency_histograms = []
+        baseline_impression_vector = []
+        for i in range(p):
+            if not i in single_pub_points.keys():
```
chenweiw: Same as the above, deleta ".keys()". In addition, do you prefer "if i not in" or "if not i in"?

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+            np.array(baseline_impression_vector),
+        )
+
+    @staticmethod
+    def obtain_observable_directions(
+        reach_points: List[ReachPoint], baseline_impression_vector: np.ndarray
+    ) -> np.ndarray:
+        """Obtain observable_directions as an input of MultivariateMixedPoissonOptimizer.
+
+        As mentioned in the description of , the observable_directions are obtain
```
chenweiw: The first sentence is incomplete....

obtain -> obtained

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+                self.optimizer.fit()
+                break
+            except Exception as inst:
+                # There is a tiny chance of exception when cvxpy mistakenly
+                # thinks the problem is non-convex due to numerical errors.
+                # If this occurs, it is likely that we have a large number
+                # of components.  In this case, try reducing the number of
+                # components.
+                logging.vlog(1, f"Optimizer failure: {inst}")
+                self.ncomponents = int(self.ncomponents / 2)
```
chenweiw: int(self.ncomponents / 2) is equivalent to self.ncomponents // 2 when of course self.ncomponents is int.

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+        count = 0
+        while curve(probe) < target_reach:
+            probe *= 2
+            count += 1
+            if count >= 10:
+                raise ValueError(
+                    "Cannot achieve target reach.  Does the given reach curve has a upper bound?"
+                )
+        right = probe
+        left = 0 if count == 0 else int(probe / 2)
```
chenweiw: int(probe / 2) -> probe // 2

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+        # Loop termination criterion:  right - left <= 1, or either left or right hits
+        # the target reach.
+        while not (
+            curve(left) == target_reach
+            or curve(right) == target_reach
+            or right - left <= 1
+        ):
+            mid = (left + right) / 2
+            mid = int(round(mid))
```
chenweiw: lines 531-532 -> mid = left + (right - left) // 2

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+            if curve(mid) < target_reach:
+                left = mid
+            else:
+                right = mid
+        if abs(curve(left) - target_reach) < abs(curve(right) - target_reach):
+            return left
+        return right
```
chenweiw, line 539: Lines 491-539: I feel the entire function might need to be optimized.... I will discuss with you later.

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+        kplus_reaches = (
+            (self.common_universe_size * relative_kplus_reaches_from_zero[1:])
+            .round(0)
+            .astype("int32")
+        )
+        return ReachPoint(impressions, kplus_reaches)
+
+    @staticmethod
+    def backsolve_impression(
+        curve: Callable, target_reach: int, starting_impression: int
```
chenweiw: Callable[[int], int]?

jiayu-google (commented):
Many thanks for your detailed comments, Chenwei!  I will address them today.

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Good point.  In the description of `observable_directions`, I added another example of 3 pubs and also rephrased some places.  Hope it make it smoother to understand the p-publisher observation_directions.

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Sure, done.

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Good point, done.

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Added "0 <= f < = F."

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Let's use Option 2 for this research codebase.  I feel we want to minimize #error messages and just keep the important ones.

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: It's a class name.  See the imports, it's in another place of codebase.

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Same

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Acknowledged.  Keeping it as it is following our previous practice in other models (such as PairwiseUnionReachSurface)

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Yep, simplified

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Nice!  Changed

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Changed "following" to "next", hope it's now grammatically clear.

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Same as above.

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: good point, `i not in` is better.

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Thanks, completed

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Thanks, changed.

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: done

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: This is definitely cleaner.  Let me still keep the original codes since (1) it's correct although long (2) if left or right somehow become float, I'm a little afraid of potential bug (without any evidence!)

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Sure.  It's likely that there're standard codes for this method, but I didn't find one that precisely fits here.  Let's treat the improvement of this method as better-to-have.

jiayu-google (commented):
Thanks Chenwei! ptal

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,666 @@
...
+                of 3 publishers.  Rows 1-3 are single-pub observable_directions,
+                rows 4-6 indicate 2-pub unions, and rows 7 indicates 3-pub union.
+
+                As of Apr 2022, all the training points from halo_simulator will
+                have a binary observable_direction (like the above example).
+                We allow non-binary observable_direction in the future.
+            frequency_histograms_on_observable_directions:  An <n * (F + 1)> matrix
+                where n = #training points, F = maximum frequency.
+                Its i-th row is the observed frequency histogram at the i-th
+                row of observable_directions, i.e., the i-th training point.
```
jiayu-google, line 86: Clarify: starts from 0, and last bucket  = F+ reach.

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,666 @@
...
+            prior_marginal_frequency_histograms:  A <p * (F + 1)> matrix where
```
jiayu-google: Consider removing this arg and including them as single pub observable directiosn

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,666 @@
...
+
+    @staticmethod
+    def obtain_observable_directions(
+        reach_points: List[ReachPoint], baseline_impression_vector: np.ndarray
+    ) -> np.ndarray:
+        """Obtain observable_directions as an input of MultivariateMixedPoissonOptimizer.
+
+        As mentioned in the description of MultivariateMixedPoissonOptimizer.__init__(),
+        the observable_directions are obtained by standardizing the impression vector of
+        each training point with the baseline_impression_vector.
```
jiayu-google: Add args. Explain the dimension of baseline_impression_vector

jiayu-google:
@riemanli @chenweiw @matthewclegg 
Refactored the codes in response to Rieman's offline comments.  More comments are most welcome!

src/models/dirac_mixture_multi_publisher_model.py
chenweiw: Sure! For your reference, I re-organize the codes in the lines 526--539 in the following:

    while left < right - 1:
        mid = left + (right - left) // 2
        if curve(mid) == target_reach:
            return mid
        elif curve(mid) < target_reach:
            left = mid
        else:
            right = mid
    if target_reach - curve(left) < curve(right) - target_reach:
        return left
    return right

jiayu-google: This is cool!  Inserted your codes, thanks a lot Chenwei!

jiayu-google (commented):
Thanks Chenwei & Rieman for the careful and constructive comments.  Please take another look when you have time.

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+                We allow non-binary observable_direction in the future.
+            frequency_histograms_on_observable_directions:  An <n * (F + 1)> matrix
+                where n = #training points, F = maximum frequency.
+                Its i-th row is the observed frequency histogram at the i-th
+                row of observable_directions, i.e., the i-th training point.
+                For each frequency histogram H, H[0] indicates the non-reach,
+                H[f] indicates the reach at frequency f for f = 1, ..., F - 1,
+                and H[F] indicates the F+ reach.
+            ncomponents:  Number of components in the Poisson mixture.
+            dilusion:  An arg to control the weights when sampling components.
```
matthewclegg: Should this be 'dilution'?

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+                    single_pub_direction_indices[pub] = k
+        if len(single_pub_direction_indices) < p:
+            raise AssertionError(
+                "Not every pub has an observable single pub direction."
+            )
+        return [
+            self.observed_pmf_matrix[single_pub_direction_indices[i]] for i in range(p)
+        ]
+
+    def in_bound_weighted_sampling(self) -> np.ndarray:
```
matthewclegg: Why is it called in_bound_weighted_sampling?  To me a more natural name might be weighted_sampling_from_marginals or weighted_sampling_from_marginals_assuming_independence.

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+
+        Returns:
+            A <C * p> matrix where C = #components and p = #pubs.
+            Each row is a component, i.e., a vector of Poisson means at all pubs.
+        """
+        marginal_pmfs = self.find_single_pub_pmfs()
+        # The following 3 lines of codes follow the same "dilusion" logic when
+        # sampling components in the single pub model.  See here:
+        # https://github.com/world-federation-of-advertisers/planning-evaluation-framework/blob/ad18cdce5959168b1620d814ba269d898c8bb117/src/models/dirac_mixture_single_publisher_model.py#L83
+        water = np.array([self.dilusion / (self.max_freq + 1)] * (self.max_freq + 1))
```
matthewclegg: It looks like what you are doing is sampling from a mixture of the true PMF and the uniform distribution, where the weight given to the uniform distribution is given by self.dilusion.  Is this correct?  If so, perhaps it would be good to update the comment?

src/models/dirac_mixture_multi_publisher_model.py
```diff
@@ -0,0 +1,632 @@
...
+            observed_arr=self.observed_pmf_matrix,
+            component_arrs=self.component_pmf_matrices,
+            distance=UnivariateMixedPoissonOptimizer.cross_entropy,
+        )
+        self.fitted = True
+
+    def predict(
+        self, hypothetical_direction: np.ndarray, customized_max_freq: int = None
+    ) -> np.ndarray:
+        """Predict frequency histogram of a hyperthetical campaign.
```
matthewclegg: hypothetical

chenweiw (approved):
LGTM

matthewclegg (changes requested):
Overall, the code is very well written!

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Sorry for my terrible spelling... Changed

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: This possibility was excluded by `UnivariateMixedPoissonOptimizer.validate_frequency_histogram(hist)` in line 100

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: yeah, agree than `in_bound` sounds weird. 

1. Now changed to `weighted_sampling_from_marginals`.
2. `in_bound` was used to emphasize that the marginal distributions are truncated by the frequency cap --- in other words, we do not sample components out of the bounds of frequency cap.  PLMK if you have another name in mind.

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Thanks for helping me phrasing this description.  Updated.

src/models/dirac_mixture_multi_publisher_model.py
jiayu-google: Thanks!

jiayu-google (commented):
Thank you Matt.  Ptal.

matthewclegg (approved):

PR #114, opened 2022-06-01
Updated m3 data design

jiayu-google:
Include CopulaDataGenerators with different params into m3_data_design.

### DIFF
--- a/src/data_generators/copula_data_set.py
+++ b/src/data_generators/copula_data_set.py
@@ -17,6 +17,7 @@
 import numpy as np
 from typing import List, Iterable, Dict
 from collections import Counter
+from scipy import linalg as splinalg
 from statsmodels.distributions.copula.copulas import (
     Copula,
     CopulaDistribution,
@@ -205,3 +206,79 @@ class CopulaDataSet(DataSet):
                 impressions[i] = impressions[i] + vids * freq_vec[i]
             num_vids += count
         return impressions
+
+
+class CopulaCorrelationMatrixGenerator:
+    @staticmethod
+    def homogeneous(p: int, rho: float) -> np.ndarray:
+        """Generate a homogeneous correlation matrix.
+
+        Args:
+            p:  Number of pubs.
+            rho:  Homogenous correlation.
+
+        Returns:
+            A p * p matrix with diagonals being 1 and all off-diagonals being rho.
+        """
+        return splinalg.toeplitz([1] + [rho] * (p - 1))
+
+    @staticmethod
+    def autoregressive(p: int, rho: float) -> np.ndarray:
+        """Generate a autoregressive correlation matrix.
+
+        Args:
+            p:  Number of pubs.
+            rho:  Homogenous correlation.
+
+        Returns:
+            A <p * p> matrix C where C[i, j] = rho^|i - j| for any i, j.
+        """
+        return splinalg.toeplitz(np.power(np.array([rho] * p), np.arange(p)))
+
+    @staticmethod
+    def random(
+        p: int, rng: np.random.Generator = np.random.default_rng(0)
+    ) -> np.ndarray:
+        """Randomly, uniformly draw a correlation matrix.
+
+        Given the dimension, all the possible correlation matrices form a space,
+        and there is a geometric measure on this space.  The geoemetric measure
+        defines a probability space.  This function uniformly draws a correlation
+        matrix from this natural probability space.
+
+        We implemented the uniform sampling algorithm in:
+            S. Ghosh, S. Henderson, "Behavior of the NORTA Method for
+            Correlated Random Vector Generation as the Dimension Increases,"
+            ACM Transactions on Modeling and Computer Simulation, Vol. 13,
+            Iss. 3, July 2003, pp. 276–294.
+
+        Args:
+            p:  Number of pubs.
+            rng:  A random number generator.
+
+        Returns:
+            A <p * p> random correlation matrix.
+        """
+        # Initliaze the correlation matrix.
+        corr = np.ones(shape=(1, 1))
+        # Increment the size of correlation matrix by one each time.
+        for k in range(2, p + 1):
+            # Sample y = r^2 from a beta distribution with alpha_1 = (k-1)/2
+            # and alpha_2 = (p + 1 - k)/2.
+            y = rng.beta((k - 1) / 2, (p + 1 - k) / 2)
+            r = np.sqrt(y)
+            # Sample a unit vector theta uniformly from the (k-1)-dimensional
+            # unit ball surface.
+            v = rng.normal(size=k - 1)
+            theta = v / np.linalg.norm(v)
+            # Set w = r * theta and set q = corr**(1/2) * w.
+            w = np.dot(r, theta)
+            q = np.dot(splinalg.sqrtm(corr), w)
+            # Incrementally create the next_corr.
+            next_corr = np.zeros((k, k))
+            next_corr[: (k - 1), : (k - 1)] = corr
+            next_corr[k - 1, k - 1] = 1
+            next_corr[k - 1, : (k - 1)] = q
+            next_corr[: (k - 1), k - 1] = q
+            corr = next_corr
+        return corr
--- a/src/data_generators/copula_data_set.py
+++ b/src/data_generators/copula_data_set.py
@@ -218,7 +218,7 @@ class CopulaCorrelationMatrixGenerator:
         measure defines a natural probability space.  This function uniformly draws
         a correlation matrix from this natural probability space.
 
-        The uniformly sampling is realized by the algorithm of:
+        The uniform sampling is realized by the algorithm of:
             D. Lewandowski, D. Kurowickaa, H. Joe, "Generating random correlation
             matrices based on vines and extended onion method," Journal of
             Multivariate Analysis, Vol. 100, Iss. 9, October 2009, pp. 1989-2001.
@@ -228,8 +228,8 @@ class CopulaCorrelationMatrixGenerator:
             eta:  A non-negative a tuning parameter.  Default eta = 1 gives a uniform
                 distribution as described above.  (Other eta can give a non-uniform
                 sampling of which the density is function of the determinant of the
-                correlation matix, and eta.  In this way we can sample higher or lower
-                correlations.)
+                correlation matix, and eta.  In this way we can tune the weights of
+                sampling stronger or weaker correlations.)
             rng:  A random number generator.
 
         Returns:
--- a/src/data_generators/copula_data_set.py
+++ b/src/data_generators/copula_data_set.py
@@ -177,7 +177,7 @@ class CopulaDataSet(DataSet):
         return impressions
 
 
-class CopulaCorrelationMatrixGenerator:    AnyFrequencyDistribution,
+class CopulaCorrelationMatrixGenerator:
     @staticmethod
     def homogeneous(p: int, rho: float) -> np.ndarray:
         """Generate a homogeneous correlation matrix.
--- a/src/data_generators/copula_data_set.py
+++ b/src/data_generators/copula_data_set.py
@@ -242,48 +242,52 @@ class CopulaCorrelationMatrixGenerator:
 
     @staticmethod
     def random(
-        p: int, rng: np.random.Generator = np.random.default_rng(0)
+        p: int,
+        eta: float = 1,
+        rng: np.random.Generator = np.random.default_rng(0),
     ) -> np.ndarray:
         """Randomly, uniformly draw a correlation matrix.
 
-        Given the dimension, all the possible correlation matrices form a space,
-        and there is a geometric measure on this space.  The geoemetric measure
-        defines a probability space.  This function uniformly draws a correlation
-        matrix from this natural probability space.
+        A p * p correlation matrix is in the (p * p) dimensional Euclidean space.
+        So, the geometric measure of (p * p) dimensional Euclidean space induces a
+        measure on the manifold of p * p correlation matrices. This geoemetric
+        measure defines a natural probability space.  This function uniformly draws
+        a correlation matrix from this natural probability space.
 
-        We implemented the uniform sampling algorithm in:
-            S. Ghosh, S. Henderson, "Behavior of the NORTA Method for
-            Correlated Random Vector Generation as the Dimension Increases,"
-            ACM Transactions on Modeling and Computer Simulation, Vol. 13,
-            Iss. 3, July 2003, pp. 276–294.
+        The uniformly sampling is realized by the algorithm of:
+            D. Lewandowski, D. Kurowickaa, H. Joe, "Generating random correlation
+            matrices based on vines and extended onion method," Journal of
+            Multivariate Analysis, Vol. 100, Iss. 9, October 2009, pp. 1989-2001.
 
         Args:
             p:  Number of pubs.
+            eta:  A non-negative a tuning parameter.  Default eta = 1 gives a uniform
+                distribution as described above.  (Other eta can give a non-uniform
+                sampling of which the density is function of the determinant of the
+                correlation matix, and eta.  In this way we can sample higher or lower
+                correlations.)
             rng:  A random number generator.
 
         Returns:
             A <p * p> random correlation matrix.
         """
-        # Initliaze the correlation matrix.
-        corr = np.ones(shape=(1, 1))
-        # Increment the size of correlation matrix by one each time.
-        for k in range(2, p + 1):
-            # Sample y = r^2 from a beta distribution with alpha_1 = (k-1)/2
-            # and alpha_2 = (p + 1 - k)/2.
-            y = rng.beta((k - 1) / 2, (p + 1 - k) / 2)
-            r = np.sqrt(y)
-            # Sample a unit vector theta uniformly from the (k-1)-dimensional
-            # unit ball surface.
-            v = rng.normal(size=k - 1)
+        # The following is a line-to-line translation of the algorithm in
+        # Section 3.2 of the paper.
+        b = (p - 2) / 2 + eta
+        u = rng.beta(b, b)
+        r12 = 2 * u - 1
+        r = np.array([[1, r12], [r12, 1]])
+        for k in range(2, p):
+            b -= 0.5
+            y = rng.beta(a=k / 2, b=b)
+            v = rng.normal(size=k)
             theta = v / np.linalg.norm(v)
-            # Set w = r * theta and set q = corr**(1/2) * w.
-            w = np.dot(r, theta)
-            q = np.dot(splinalg.sqrtm(corr), w)
-            # Incrementally create the next_corr.
-            next_corr = np.zeros((k, k))
-            next_corr[: (k - 1), : (k - 1)] = corr
-            next_corr[k - 1, k - 1] = 1
-            next_corr[k - 1, : (k - 1)] = q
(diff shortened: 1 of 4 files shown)


src/data_generators/m3_data_design.py
```diff
@@ -157,20 +166,198 @@
...
+        p: int, rng: np.random.Generator = np.random.default_rng(0)
+    ) -> np.ndarray:
+        """Randomly, uniformly draw a correlation matrix.
+
+        Given the dimension, all the possible correlation matrices form a space,
+        and there is a geometric measure on this space.  The geoemetric measure
+        defines a probability space.  This function uniformly draws a correlation
+        matrix from this natural probability space.
+
+        We implemented the uniform sampling algorithm in Section 3.2 of:
```
chenweiw: Delete "Section 3.2 in" because it is used to locate the algorithm in the other reference published in 2009.

src/data_generators/m3_data_design.py
```diff
@@ -157,20 +166,198 @@
...
+
+        Returns:
+            A <p * p> random correlation matrix.
+        """
+        # Initliaze the correlation matrix.
+        corr = np.ones(shape=(1, 1))
+        # Increment the size of correlation matrix by one each time.
+        for k in range(2, p + 1):
+            # Sample y = r^2 from a beta distribution with alpha_1 = (k-1)/2
+            # and alpha_2 = (d-k)/2.
```
chenweiw: In the description, change "d" to "p+1" for alpha_2.

src/data_generators/m3_data_design.py
```diff
@@ -157,20 +166,198 @@
...
+                },
+                "random_generator": np.random.default_rng(RNG.integers(1e9)),
+            },
+        )
+        for ratio in [0.25, 0.75]
+        for df in [2, 10]
+        for rho in [-0.5, 0.5]
+    ]
+)
+
```
chenweiw, line 283: I personally feel the list generation per the lines 253 -- 355 are too long, although the codes are easier to read. Since there are 5 sub-lists are concatenated, could we add some description to each sub-list? To do so, we can generate sub-lists like a = []. b = [], c = [], d = [], e = [] with some description, and then the final list is a + b + c + d + e, which makes the codes easier to read.

src/data_generators/tests/copula_data_set_test.py
```diff
@@ -58,6 +61,11 @@ def test_to_impressions(self):
         np.testing.assert_equal(res[0], expected[0])
         np.testing.assert_equal(res[1], expected[1])
```
chenweiw: Can lines 61--62 be combined into one line, i.e., np.testing.assert_equal(res, expected)?

src/data_generators/tests/copula_data_set_test.py
```diff
@@ -141,6 +145,89 @@ def test_fully_negatively_correlated_copula(self):
         self.assertFalse((1, 1) in res)
         self.assertFalse((2, 2) in res)
 
+    # When using statsmodels.distributions.copula for more than 2 pubs, there is
+    # "UserWarning: copulas for more than 2 dimension is untested."
```
chenweiw, line 139: WOW! This is unbelievable....

src/data_generators/tests/copula_data_set_test.py
```diff
@@ -141,6 +145,89 @@ def test_fully_negatively_correlated_copula(self):
...
+                res = dataset.frequency_vectors_sampled_distribution
+                for key in product(*tuple([(1, 2)] * num_pubs)):
+                    self.assertTrue(key in res)
+                    self.assertAlmostEqual(res[key] / 100, 1, delta=0.5)
+
+    def test_fully_positively_correlated_copula_with_more_than_two_pubs(self):
+        # Note: fully negatively correlated cases do not exist with more than 2 pubs.
+        for num_pubs in [3, 4]:
+            correlation_matrix = np.ones((num_pubs, num_pubs)) - 1e-9
+            np.fill_diagonal(correlation_matrix, 1)
```
chenweiw: Lines 209--210 can be replaced by cor_mat = linalg.toeplitz([1] + [1 - 1e-9] * (num_pubs - 1)) following the same type of codes in Line 158.

src/data_generators/tests/copula_data_set_test.py
```diff
@@ -141,6 +145,89 @@ def test_fully_negatively_correlated_copula(self):
...
+                    universe_size=100 * 2 ** num_pubs,
+                    random_generator=np.random.default_rng(0),
+                )
+                res = dataset.frequency_vectors_sampled_distribution
+                for key in product(*tuple([(1, 2)] * num_pubs)):
+                    self.assertTrue(key in res)
+                    self.assertAlmostEqual(res[key] / 100, 1, delta=0.5)
+
+    def test_fully_positively_correlated_copula_with_more_than_two_pubs(self):
+        # Note: fully negatively correlated cases do not exist with more than 2 pubs.
```
chenweiw: Right, or we can consider some "hybrid" cases like, e.g., for 3 pubs, pub 1 and pub 2 are fully positively correlated and they are both fully negatively correlated with pub 3. You can decide if it deserves a test or not.

src/data_generators/m3_data_design.py
```diff
@@ -52,8 +53,8 @@
 # The following are the parameter sets that are varied in this data design.
 # The latin hypercube design constructs a subset of the cartesian product
 # of these parameter settings.
-NUM_PUBLISHERS = [1, 2, 5, 10, 20, 100]
-LARGEST_PUBLISHER = [int(1e4), int(1e5), int(1e6), int(1e7)]
+NUM_PUBLISHERS = [2, 5, 10, 20]
+LARGEST_PUBLISHER = [int(1e5), int(1e6), int(1e7)]
```
jiayu-google: replace 1e7 with 1e4.

jiayu-google: another option: reach = 1e5, 2e5, 3e5.  might just have 1 dataset with size 1e6 to confirm (which may be put in a separate stree_data_design.py).

src/data_generators/m3_data_design.py
```diff
@@ -52,8 +53,8 @@
...
 PUBLISHER_RATIOS = [1, 0.5, 0.3, 0.1, 0.01]
```
jiayu-google: eliminate 0.01.

src/data_generators/m3_data_design.py
jiayu-google: thanks, good suggestion

src/data_generators/tests/copula_data_set_test.py
jiayu-google: I didn't know np.testing can be applied to list of list (as the `expected` in line 57).  This is good to know, thanks!

src/data_generators/tests/copula_data_set_test.py
jiayu-google: Good suggestion, let me add this test case.

src/data_generators/tests/copula_data_set_test.py
jiayu-google: Thanks!

chenweiw (approved):
LGTM

src/data_generators/m3_data_design.py
```diff
@@ -91,32 +92,40 @@
...
-OVERLAP_GENERATORS = [
+RNG = np.random.default_rng(0)
+OVERLAP_GENERATORS_INDEPENDENT_GIVEN_UNIVERSE_SIZE = [
     GeneratorParameters("FullOverlap", DataSet, {}),
     GeneratorParameters(
         "Independent",
         IndependentOverlapDataSet,
-        {"largest_pub_to_universe_ratio": 0.9, "random_generator": 1},
+        {
+            "largest_pub_to_universe_ratio": 0.9,
```
matthewclegg, line 107: Is there a relationship between this value and the values given in PUBLISHER_RATIOS at line 58?

src/data_generators/m3_data_design.py
```diff
@@ -157,19 +166,200 @@
...
             "correlated_sets": CorrelatedSetsOptions.one,
             "shared_prop": 0.75,
-            "random_generator": 8,
+            "random_generator": np.random.default_rng(RNG.integers(1e9)),
         },
     ),
+]
+
+
+class CopulaCorrelationMatrixGenerator:
```
matthewclegg: Move this to a separate file?

src/data_generators/m3_data_design.py
matthewclegg: +1

src/data_generators/m3_data_design.py
```diff
@@ -157,19 +166,200 @@
...
+        return splinalg.toeplitz(np.power(np.array([rho] * p), np.arange(p)))
+
+    @staticmethod
+    def random(
+        p: int, rng: np.random.Generator = np.random.default_rng(0)
+    ) -> np.ndarray:
+        """Randomly, uniformly draw a correlation matrix.
+
+        Given the dimension, all the possible correlation matrices form a space,
+        and there is a geometric measure on this space.  The geoemetric measure
```
matthewclegg: Could you further explain what you mean by geometric measure?

src/data_generators/m3_data_design.py
```diff
@@ -157,19 +166,200 @@
...
+            p:  Number of pubs.
+            rho:  Homogenous correlation.
+
+        Returns:
+            A <p * p> matrix C where C[i, j] = rho^|i - j| for any i, j.
+        """
+        return splinalg.toeplitz(np.power(np.array([rho] * p), np.arange(p)))
+
+    @staticmethod
+    def random(
```
matthewclegg: Is there a unit test for this?

src/data_generators/m3_data_design.py
```diff
@@ -157,19 +166,200 @@
...
+            next_corr[k - 1, k - 1] = 1
+            next_corr[k - 1, : (k - 1)] = q
+            next_corr[: (k - 1), k - 1] = q
+            corr = next_corr
+        return corr
+
+
+# Following this design that was reviewed by WFA:
+# https://docs.google.com/document/d/1pRA_fc0RbhRVUPsxbmUDmcrvNRwpNILpO-CGZtG5gYI/edit#
+OVERLAP_GENERATORS_COPULA_GAUSSIAN_HOMO = [
```
matthewclegg: Change this to OVERLAP_GENERATORS_COPULA_GAUSSIAN_HOMOGENEOUS

src/data_generators/m3_data_design.py
```diff
@@ -157,19 +166,200 @@
...
+                "generator": CopulaCorrelationMatrixGenerator.autoregressive,
+                "kwargs": {"rho": rho},
+            },
+            "random_generator": np.random.default_rng(RNG.integers(1e9)),
+        },
+    )
+    for ratio in [0.25, 0.75]
+    for rho in [-0.5, -0.25, 0.25, 0.5]
+]
+OVERLAP_GENERATORS_COPULA_GAUSSIAN_RAND = [
```
matthewclegg: Change this to OVERLAP_GENERATORS_COPULA_GAUSSIAN_RANDOM

src/data_generators/m3_data_design.py
```diff
@@ -157,19 +166,200 @@
...
+                "generator": CopulaCorrelationMatrixGenerator.random,
+                "kwargs": {"rng": np.random.default_rng(seed)},
+            },
+            "random_generator": np.random.default_rng(RNG.integers(1e9)),
+        },
+    )
+    for ratio in [0.25, 0.75]
+    for seed in [1, 2, 3, 4]
 ]
+OVERLAP_GENERATORS_COPULA_T_HOMO = [
```
matthewclegg: OVERLAP_GENERATORS_COPULA_T_HOMOGENEOUS

src/data_generators/m3_data_design.py
jiayu-google: largest_pub_to_universe_ratio = largest_pub_reach / universe_size.
PUBLISHER_RATIOS = smallest_pub_reach / largest_pub_reach.
So, I think they are unrelated.  Together with `LARGEST_PUBLISHER` in line 56 they determine all the sizes. `"largest_pub_to_universe_ratio"` is used only when the universe size is needed.

src/data_generators/m3_data_design.py
jiayu-google: Thanks. Moved to copula_data_set.py

src/data_generators/m3_data_design.py
jiayu-google: There was a basic test which checked if the generated matrices are positive definite and thus a valid correlation matrix.  I will add another test to check the random-uniformity at least when p = 2, after confirming the definition of uniformity with Chenwei.

src/data_generators/m3_data_design.py
jiayu-google: A p * p correlation matrix is in the (p * p) dimensional Euclidean space.  So, the geometric measure of (p * p) dimensional Euclidean space induces a measure on the manifold of p * p correlation matrices.  
This was my interpretation from the abstract + introduction of the paper, and I regret not having checked the math details.  **Now I doubt if the generated matrices are indeed uniform in this sense.**

@chenweiw Could you please chime in?  Per discussion with Chenwei, the paper is not very easy to read.  Let's spend more time reading it, OR do some experiments to confirm the definition of uniformity. WDYT?

src/data_generators/m3_data_design.py
jiayu-google: Thanks, done

jiayu-google (commented):
Thanks Matt!

@chenweiw Per Matt's comments, I think we need to double check the definition of the uniformity of the randomly generated correlation matrices.  tty today.

src/data_generators/m3_data_design.py
matthewclegg: ok, thanks for explaining that.  Would it be helpful to add a comment to the code so that others don't have this same question?

matthewclegg (approved):

src/data_generators/m3_data_design.py
jiayu-google: Sure, added!

jiayu-google (commented):
Thanks to @chenweiw 's huge help, ensured the uniformity of the generated correlation matrix.
More uniformity validation was done offline in Chenwei's Colab.
@matthewclegg  @chenweiw  Please take another look.

src/data_generators/copula_data_set.py
```diff
@@ -174,3 +175,83 @@ def to_impressions(frequency_vectors: List[np.ndarray]) -> List[List[int]]:
...
+    ) -> np.ndarray:
+        """Randomly, uniformly draw a correlation matrix.
+
+        A p * p correlation matrix is in the (p * p) dimensional Euclidean space.
+        So, the geometric measure of (p * p) dimensional Euclidean space induces a
+        measure on the manifold of p * p correlation matrices. This geoemetric
+        measure defines a natural probability space.  This function uniformly draws
+        a correlation matrix from this natural probability space.
+
+        The uniformly sampling is realized by the algorithm of:
```
chenweiw: uniformly -> uniform

chenweiw: As onion method is not the only one algorithm that the paper introduced, do you wanna identify it in Section 3.2 here (besides the line 239) or add the name of "onion method"?

src/data_generators/copula_data_set.py
```diff
@@ -174,3 +175,83 @@ def to_impressions(frequency_vectors: List[np.ndarray]) -> List[List[int]]:
...
+            D. Lewandowski, D. Kurowickaa, H. Joe, "Generating random correlation
+            matrices based on vines and extended onion method," Journal of
+            Multivariate Analysis, Vol. 100, Iss. 9, October 2009, pp. 1989-2001.
+
+        Args:
+            p:  Number of pubs.
+            eta:  A non-negative a tuning parameter.  Default eta = 1 gives a uniform
+                distribution as described above.  (Other eta can give a non-uniform
+                sampling of which the density is function of the determinant of the
+                correlation matix, and eta.  In this way we can sample higher or lower
```
chenweiw: I think you mean "stronger or weaker" correlation?

src/data_generators/tests/copula_data_set_test.py
```diff
@@ -130,6 +135,224 @@ def test_fully_negatively_correlated_copula(self):
...
+            (cor_mat[0, 1], cor_mat[0, 2], cor_mat[1, 2]) for cor_mat in cor_mats
+        ]
+        polar_tuples = [
+            to_polar_coordinates(*correlation_tuple)
+            for correlation_tuple in correlation_tuples
+        ]
+
+        # 1. Test the uniformity of the radius.  Because of the positive definitiveness
+        # constraint, not all values of the radius are possible.  It can be proved that
+        # When (x, y, z) has a radius no greater than 0.5, the correlation matrix of the
```
chenweiw, line 307: Does 0.5 come from the lower bound of the correlation coefficient for the homogeneous setting that of course gives this very conservative radius bound (because based on our investigation it seems radius < roughly 0.8 is generally ok)?

src/data_generators/tests/copula_data_set_test.py
```diff
@@ -130,6 +135,224 @@ def test_fully_negatively_correlated_copula(self):
...
+                bins=4,
+                range=(0, 0.5),
+            )[0],
+            dtype=float,
+        )
+        radius_cube_frequencies /= sum(radius_cube_frequencies)
+        np.testing.assert_almost_equal(radius_cube_frequencies, [1 / 4] * 4, decimal=1)
+
+        # 2. Test the uniformity of the phi angle.  Another necessary condition for the
+        # uniformity of the whole correlation matrix is that phi is uniform in [-pi, pi].
```
chenweiw, line 325: Due to phi = arctan(./.), should be phi uniform in (-pi/2, pi/2)?

chenweiw (commented):
Made some minor comments on the onion method and its testing.

matthewclegg (approved):

src/data_generators/copula_data_set.py
jiayu-google: Done

jiayu-google: Good point.  IMO there are pros and cons of renaming it.  "Onion method" is not a well known concept, so names like `random_onion` may introduce confusion.  For now, I'd like to keep the name as simple as `random`, and we can always change the name later if we want to implement another random sampling method.

src/data_generators/copula_data_set.py
jiayu-google: yep.  Rephrased following your suggestion.

src/data_generators/tests/copula_data_set_test.py
jiayu-google: Yeah the limit should be around 0.8.   I was able to *prove* that 0.5 works by some loose inequalities.  To prove 0.8 we need tighter inequalities which I didn't try.

src/data_generators/tests/copula_data_set_test.py
jiayu-google: You are right, and that's why I prefer `arctan2`.  `arctan2(x, y)` recognizes the signs of x, y and finds an angle in [0, 2 \pi).   Thinking about `phi` as the angle in the 2d polar coordinates, its domain is [0, 2\pi).

jiayu-google (commented):
Thanks for the ideas of random sampling and the careful review, Chenwei.

PR #116, opened 2022-06-09
Add data generator for use with testing fake EDPs.

matthewclegg:


matthewclegg:
The purpose of this code is to simplify generating synthetic data sets that can be used for testing the performance of Halo.  As input, this program expects a CSV file specifying various data about the EDP's.  See the comments at the top of the file for definitions of the various columns.  Following is an example input file:

```
publisher_id,mc_id,gender,age_group,social_grade,date,mean,std,cardinality
1,mc-5,M,18-24,ABC1,2022/01/02,2.0,3.0,3
1,mc-5,M,25-34,ABC1,2022/01/02,2.0,3.0,5
2,mc-5,M,18-24,ABC1,2022/01/02,2.0,3.0,3
```

Assuming that this file is stored as `~/dataconfig.csv`, the following command could be used to generate a synthetic data set:

```
python3 edp_data_generator.py --output=~/synthetic-data.csv --data_design_config=~/dataconfig.csv 
```

This command would be executed from the directory `planning-evaluation-framework/src/data_generators`.

Please note that this CL builds on `jiayu_copula_data_set`.

src/data_generators/edp_data_generator.py
```diff
@@ -0,0 +1,238 @@
+# Copyright 2021 The Private Cardinality Estimation Framework Authors
```
jiayu-google: 2022

src/data_generators/edp_data_generator.py
```diff
@@ -0,0 +1,238 @@
...
+
+class EdpDataGenerator:
+    """Generates a synthetic data for use in testing Halo."""
+
+    def __init__(
+        self,
+        data_design_config: str,
+        random_seed: int = 1,
+    ):
+        """Constructor for SyntheticDataGenerator.
```
jiayu-google: EdpDataGenerator

src/data_generators/edp_data_generator.py
```diff
@@ -0,0 +1,238 @@
...
+          A PublisherData object representing the list of impressions associated
+          to this demo group.
+        """
+        mean = self._data_design_config["mean"].iloc[i]
+        std = self._data_design_config["std"].iloc[i]
+        n = self._data_design_config["cardinality"].iloc[i]
+        if mean > std:
+            raise ValueError(
+                "Invalid values for mean and std. "
+                "Mean must be less than std. "
```
jiayu-google: I think the requirement is "mean - 1 must be less than std ** 2."  Reason:
1. Line 146.
2. For Poisson, mean = variance, not std.  So for GP, mean < variance.  We shifted GP by 1, so mean - 1 < variance.

src/data_generators/edp_data_generator.py
```diff
@@ -0,0 +1,238 @@
...
+output file are as follows:
+
+    Publisher ID,Event ID,Sex,Age Group,Social Grade,Date,Complete,VID
+
+For each combination of (publisher_id, mc_id, gender, age_group,
+social_grade, date), approximately cardinality many rows are generated
+in the output.  The distribution of impressions per viewer follows a
+negative binomial distribution with specified mean and standard
+deviation.  If multiple publishers are specified in the input, then
+the overlap is determined using a Gaussian copula where the data is
```
jiayu-google: maybe just say "Independent copula", instead of "Gaussian copula where the data is assumed to be independent"?  you didn't import Gaussian copula below

src/data_generators/tests/edp_data_generator_test.py
```diff
@@ -0,0 +1,117 @@
...
+        expected_data = pd.DataFrame(
+            {
+                "Publisher ID": [1, 1, 2, 2, 2, 1, 1],
+                "Event ID": [0, 1, 2, 3, 4, 5, 6],
+                "Sex": ["M", "M", "M", "M", "M", "F", "F"],
+                "Age Group": ["18-24"] * 5 + ["25-34"] * 2,
+                "Social Grade": ["ABC1"] * 5 + ["ABC2"] * 2,
+                "Date": ["2022/01/02"] * 5 + ["2022/01/03"] * 2,
+                "Complete": ["1"] * 7,
+                "VID": list([0, 3, 1, 2, 3, 5, 4]),
```
jiayu-google, line 109: Could you help me understand this result?  Here pub 2 reached three vids 1, 2, 3.  But above, the cardinality was set as 2.

jiayu-google (changes requested):
Thanks Matt! Left a few comments.

src/data_generators/tests/edp_data_generator_test.py
matthewclegg: In my view, this is actually a bug in GaussianCopulaDataSet.  By the way, to be consistent with IndependentOverlapDataSet and SequentiallyCorrelatedOverlapDataSet, this should be called GaussianCopulaOverlapDataSet, and it should inherit from OverlapDataSet.  In any event, it would be preferable if GaussianCopula(Overlap)DataSet guaranteed that the cardinality of each publisher in the output set is identical to the cardinality of the corresponding publisher in the input data set.  It would be even better if it could guarantee that the marginal distributions are the same, but I don't see an easy way to guarantee that.

jiayu-google: Thanks! Makes great sense.  Two initial responses:
1.  Good catch, the copula data set only has **approximate** single pub agreement, because the data set is obtained by random sampling.  Like you said, I didn't see an easy way either to guarantee exact agreement.
2. I've thought about making CopulaDataSet a subclass of OverlapDataSet and it didn't seem very easy to refactor `OverlapDataSet`.   I regret not having a great design of `OverlapDataSet` a year ago.
Anyway, the discrepancy here is well explained, so I'm approving your PR.  For the above 2 issues, LMK if any idea pops up in your mind.

jiayu-google (approved):
LGTM.

src/data_generators/tests/edp_data_generator_test.py
matthewclegg: Thanks for reviewing and approving the code.  I think it would be easy to modify CopulaDataSet so that the single publisher data sets in the output have the same cardinality as in the input.  This could be done as follows:

1.  Generate enough random samples such that the total cardinality of each publisher in the output set equals or exceeds the total cardinality in the input set.  
2. For each publisher where the cardinality in the output set exceeds the cardinality in the input set, repeat the following:  Choose a user at random and delete all of the impressions of that user in the given publisher.

The joint distribution of the resulting data set will deviate very slightly from the true distribution defined by the Gaussian copula.

jiayu-google: Good idea:) Will modify PR 108 tomorrow

jiayu-google: From today's meeting I think I have some more urgent AIs, for example, enabling the check of single pub k+ agreement in experimental_trial.  So, I'll defer the improvement of `CopulaDataSet`, plmk if there's any objection.  (If by any chance you have bandwidth to realize this improvement, it'll be awesome!)

